PR #103, opened 2015-08-07
Dd 1002 db migration lock

elvisciotti:


src/AppBundle/Command/MigrationsMigrateLockCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * release lock
+     * 
+     * @param type $output
+     */
+    private function releaseLock($output)
+    {
+        $output->writeln('Lock released.');
+        
+        return $this->getRedis()->del(self::LOCK_KEY);
```
munhitsu: what happens if we delete non existing key? - will it work or raise exception?

munhitsu: clarified that it's safe - no exception

PR #160, opened 2016-01-20
Report fixes

elvisciotti:
fix missing transactions (migration unit-tsted)
    fix order
    fix phpunit tests
    phpunit tests not clear data before each tests using pg import (more independent runs)


elvisciotti:
- fix missing transactions (migration unit-tsted)
- fix order
- fix phpunit tests
- phpunit tests not clear data before each tests using pg import (more independent runs)

PR #167, opened 2016-02-01
dd-1336-user-not-found-redirect user not found return code 419 for re…

elvisciotti:
…direction to login page


elvisciotti:
to merge tomorrow after Kate's demo

PR #226, opened 2016-12-21
OPGOPS-1677 Inital dev setup of digideps

carinadigital:


elvisciotti:
@carinadigital  is the "X-Request-Id" logic ported over ? we need that for logstash

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.service.opg.digital/opguk/php-fpm
+FROM registry.service.opg.digital/opguk/php-fpm:0.1.207
```
brettminnie: Why are we locking to a version here? I was under the impression we removed this so that when built latest stable would always be pulled in to make sure security updates etc where applied

PR #280, opened 2017-03-08
DDPB-1019-pa-team-name store/load PA team name [TO DISCUSS]

elvisciotti:


elvisciotti:
@slizzio  approach that lets easily save the team name as part of the form, simplifying the client logic, and only one endpoint

app/DoctrineMigrations/Version125.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE dd_team ALTER team_name DROP NOT NULL');
```
slizzio: why would we want to create a team with no name?

src/AppBundle/Entity/User.php
```diff
@@ -258,8 +258,9 @@ class User implements UserInterface
      */
     public function __construct()
     {
-        $this->clients = new \Doctrine\Common\Collections\ArrayCollection();
+        $this->clients = new ArrayCollection();
         $this->password = '';
+        $this->teams = new ArrayCollection();
```
slizzio: This constructor code always creates an empty object. This might be one reason why prevous attempts at doctrine persistence wasnt working...? A constructor should be capable of creating new and existing User objects: new User(); or new User($clients, $teams);
what about something like this: 
     public function __construct($clients = [], $teams = [])
      {
          $this->clients = new ArrayCollection($clients);
          $this->teams = new ArrayCollection($teams);
          ...
By specifying rules in the entity, it keeps business logic in the backend and *anything* that creates a User needs to adhere to the contract that the object specifies. In this case, to create a user I expect a list of clients and teams (although they can be empty)

src/AppBundle/Entity/User.php
```diff
@@ -956,6 +957,22 @@ public function setJobTitle($jobTitle)
         return $this;
     }
 
+    /**
+     * Return Id of the client (if it has details)
```
slizzio: Return Team name

src/AppBundle/Controller/UserController.php
```diff
@@ -353,6 +353,14 @@ private function populateUser(EntityDir\User $user, array $data)
             $user->setLastLoggedIn(new \DateTime($data['last_logged_in']));
         }
 
+        if (!empty($data['pa_team_name'])) {
+            if ($user->getTeams()->isEmpty()) {
+                $user->addTeam(new EntityDir\Team());
+            }
+            $team = $user->getTeams()->first()->setTeamName($data['pa_team_name']);
+            $this->persistAndFlush($team);
+        }
```
slizzio: Id pass the team name as a constructor argument to the Team entity (and make it optional if there is a use case for having teams with no name). Keeps business logic in api layer.

elvisciotti: changed

src/AppBundle/Entity/User.php
elvisciotti: my change was only initialising the teams, otherwise null and I couldn't safely call "isEmpty"

PR #300, opened 2017-05-10
Ddpb 1201 - fix data

elvisciotti:


iterative-it:
Excellent work! I assume that you had to look at the prod data you extracted to diagnose this? Will this script need re running each time or has the underlying issue been fixed?

PR #304, opened 2017-05-16
DDPB-1170

iterative-it:
Generating the correct error codes when a user's email is changed to an already existing one.

### DIFF
--- a/src/AppBundle/Controller/UserController.php
+++ b/src/AppBundle/Controller/UserController.php
@@ -98,8 +98,6 @@ class UserController extends RestController
             $userService->editUser($originalUser, $user);
         };
 
-        $this->getEntityManager()->flush($user);
-
         return ['id' => $user->getId()];
     }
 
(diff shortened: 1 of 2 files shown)


src/AppBundle/Controller/UserController.php
```diff
@@ -81,17 +81,21 @@ public function update(Request $request, $id)
...
         if ($loggedInUser->isPaAdministrator()) {
-            if (empty($user->getRoleName())) {
-                $user->setRoleName(EntityDir\User::ROLE_PA_TEAM_MEMBER);
-            }
+            $userService->editPaUser($originalUser, $user);
+        } else {
+            $userService->editUser($originalUser, $user);
         };
 
         $this->getEntityManager()->flush($user);
```
slizzio: Should move this to the service too

src/AppBundle/Service/UserService.php
```diff
@@ -58,4 +58,33 @@ public function addPaUser(User $loggedInUser, User $userToAdd, $data)
             }
         }
     }
+
+    public function editUser(User $originalUser, User $userToEdit)
```
slizzio: docblock

PR #306, opened 2017-05-17
DDPB-1215 user/get-by-token return strictly-needed groups only

elvisciotti:


elvisciotti:
@slizzio  @iterative-it  this endpoint recursively returned all the data, ending up in failing on the client side when dehydrating MoneyTransactionShort.report that missed the JMS type. 
I fixed it by returning only the needed data, as other endpoints do.
I fixed also the dehydration bit on the client (see https://github.com/ministryofjustice/opg-digi-deps-client/pull/658/files) in case we need to return recursively anything included that entity (only the rare case where a groups is normally defined)

PR #314, opened 2017-05-30
Ddpb 1239 report type csv updated

elvisciotti:


elvisciotti:
- Client now required to construct Report
- report.type logic moved to single place (ReportService)
- test more "unit" for the report type logic
- "a3" is integrated but need to better integrate/clean up things after the changes from master

PR #316, opened 2017-06-01
DDPB-1239: Set report type based on last year's report if casrec chec…

iterative-it:
…k fails

### DIFF
--- a/src/AppBundle/Service/ReportService.php
+++ b/src/AppBundle/Service/ReportService.php
@@ -53,7 +53,7 @@ class ReportService
             return;
         }
 
-        if($report instanceof Report)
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/ReportService.php
```diff
@@ -53,6 +53,12 @@ public function setReportTypeBasedOnCasrec(Report $report)
             return;
         }
 
+        if($report instanceof Report and $report->getType() !== null)
```
elvisciotti: the method type hints "(Report $report)", so if a non-report is passed to the method, it'll throw a fatal exception and die. You could omit "$report instanceof Report and " if you want

PR #319, opened 2017-06-05
DDPB-807-tests more tests for PA Service

elvisciotti:


elvisciotti:
@slizzio small one, when you have time

PR #321, opened 2017-06-08
DDPB-807-fix-dob dob fix

elvisciotti:


tests/AppBundle/Service/PaServiceStaticTest.php
```diff
@@ -9,28 +9,36 @@ class PaServiceStaticTest extends \PHPUnit_Framework_TestCase
...
+            ['20-MAR-2003', '2003-03-20', '20'],
+            ['29-MAY-2013', '2013-05-29', '20'],
+            ['07-OCT-2016', '2016-10-07', '20'],
+            ['07-OCT-1945', '1945-10-07', '20'], //third param ignored if full year is given
 
             // invalid date
-            ['07-xxx-2016', false],
+            ['07-xxx-2016', false, '20'],
+            ['32-MAY-2016', false, '20'],
+            ['32-MAY-2016', false, '20'],
```
slizzio: could we have a test with 1 and 3 number years? As data is coming from CSV, theres a chance data may be edited incorrectly

PR #322, opened 2017-06-09
Ddpb 808 edit client details

elvisciotti:


src/AppBundle/Entity/Client.php
```diff
@@ -554,6 +554,19 @@ public function getUnsubmittedReports()
...
+     * @JMS\VirtualProperty
+     * @JMS\Type("AppBundle\Entity\Report\Report")
+     * @JMS\SerializedName("current_report")
+     * @JMS\Groups({"current-report"})
+     *
+     * @return Report|null
+     */
+    public function getCurrentReport()
+    {
+        return $this->getUnsubmittedReports()->first() ?: null;
```
slizzio: If the order of reports in the method above is altered then this will start to fail. Can we put through a sort function to ensure that the current report always returns the current report?

elvisciotti: the method uses "getUnsubmittedReports", that only consider the unsubmitted ones, and all the client only have one report active.

The inconsistency is that the business logic doesn't allow the user to create more than one unsubmitted report, but since that's not enforced in the model, in the past users managed to create more than one by re-submitted the page by mistake. 
I think we should firstly move the report creation into a service, then adding this check (only one unsubmitted a time), and rename the method to "getUnsubmittedReports->getUnsubmittedReport" (as there only is one)

slizzio: Your solutions works for me. As a quick fix would something like this be relevant?
usort(
            $this->getUnsubmittedReports(),
            function ($a, $b) {
                return strtotime($b['submitted_date']) - strtotime($a['submitted_date']);
            }
        ); to reorder them.. then take the first record.

slizzio: Maybe create a new ticket for this if it needs to be merged?

PR #330, opened 2017-06-21
Rest jms check

elvisciotti:


elvisciotti:
- REST listener requires a JMS group if an entity is detected in the response. If not, a clear exception is thrown (this would have made easier spotting the edit note save crashing). Exception messages of course skip this check
- groups set for all the endpoints (When missing). I've tested they are the right ones. That will avoid  potential recursion and crashing (happened once with user 103 activation)
- PUT endpoints aligned to return the changed record ID (default behaviour)

src/AppBundle/EventListener/RestInputOuputFormatter.php
```diff
@@ -111,6 +123,20 @@ private function arrayToResponse($data, Request $request)
...
+        if (is_array($object)) {
+            foreach($object as $subObject) {
+                if ($this->containsEntity($subObject)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        return is_object($object) && strpos(get_class($object), 'Entity') !== false;
```
slizzio: Is there not an interface we can use here? If we're using strpos we may as well var_export the whole object and call strpos(), and remove the loop

src/AppBundle/EventListener/RestInputOuputFormatter.php
```diff
@@ -76,21 +76,28 @@ public function requestContentToArray(Request $request)
...
+     * 
+     *
      *
+     * @param $data
+     * @param Request $request
+     * @param bool $groupsCheck
      * @return Response
      */
-    private function arrayToResponse($data, Request $request)
+    private function arrayToResponse(array $data, Request $request, $groupsCheck = false)
```
slizzio: should we have the checks on by default?

elvisciotti: changed

src/AppBundle/EventListener/RestInputOuputFormatter.php
elvisciotti: I thought about that but var_export could have to deal with a very big object (hope not a recursion) therefore crashing the API

PR #331, opened 2017-06-22
DDPB-1280 do not change report type for submitted report on PA CSV up…

elvisciotti:
…load

### DIFF
--- a/src/AppBundle/Service/PaService.php
+++ b/src/AppBundle/Service/PaService.php
@@ -280,7 +280,8 @@ class PaService
         $reportType = EntityDir\CasRec::getTypeBasedOnTypeofRepAndCorref($row['Typeofrep'], $row['Corref']);
         $report = $client->getReportByEndDate($reportEndDate);
         if ($report) {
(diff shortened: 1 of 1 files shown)


src/AppBundle/Service/PaService.php
```diff
@@ -280,7 +280,8 @@ private function createReport(array $row, EntityDir\Client $client, EntityDir\Us
         $reportType = EntityDir\CasRec::getTypeBasedOnTypeofRepAndCorref($row['Typeofrep'], $row['Corref']);
         $report = $client->getReportByEndDate($reportEndDate);
         if ($report) {
-            if ($report->getType() != $reportType) {
+            // change report type if it's not already set AND report is not yet submitted
+            if ($report->getType() != $reportType && !$report->getSubmitted()) {
                 $this->log('Changing report type');
```
slizzio: Can we have something like "Changing report type from <old> to <new> for report XX". It'll get lost otherwise and doesn't give any useful information.

PR #334, opened 2017-06-23
DDPB-1321 Report ctor refactor + dependencies and test updated

elvisciotti:


src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -25,24 +29,20 @@ public function addAction(Request $request)
...
         $this->validateArray($reportData, [
             'start_date' => 'notEmpty',
             'end_date' => 'notEmpty',
         ]);
 
-        // add other stuff
-        $report->setStartDate(new \DateTime($reportData['start_date']));
-        $report->setEndDate(new \DateTime($reportData['end_date']));
+        // report type is taken from CASREC. In case that's not available (shouldn't happen unless casrec table is dropped), use a 102
+        $reportType = $this->get('opg_digideps.report_service')->getReportTypeBasedOnCasrec($client) ?: Report::TYPE_102;
```
slizzio: Defaulting to a 102 should remain in the service method. Business logic in a single place

src/AppBundle/Entity/Report/Report.php
```diff
@@ -207,12 +207,51 @@ class Report
...
+        if (!in_array($type, [self::TYPE_102, self::TYPE_103, self::TYPE_104])) {
+            throw new \InvalidArgumentException("$type not a valid report type");
+        }
+        $this->type = $type;
         $this->client = $client;
+        $this->startDate = new \DateTime($startDate->format('Y-m-d'));
+        $this->endDate = new \DateTime($endDate->format('Y-m-d'));
+
+        // date interval check
+        if ((int)$startDate->diff($endDate)->format('%a') > 365) { // TODO reduce to ">=" once the db has been migrated
```
slizzio: consider leap years?

src/AppBundle/Service/ReportService.php
```diff
@@ -41,59 +42,60 @@ public function findById($id)
...
-
-        // clone assets
-        foreach ($report->getAssets() as $asset) {
+        if (!$oldReport->getSubmitted()) {
+            throw new \RuntimeException("Can't create a new year report based on an unsubmitted report");
+        }
+
+        $client = $oldReport->getClient();
+
+        $startDate = clone $oldReport->getEndDate();
```
slizzio: do we need to clone the entire object? 
$startDate = $oldReport->getEndDate() should do it ?

src/AppBundle/Service/ReportService.php
```diff
@@ -41,59 +42,60 @@ public function findById($id)
...
+        $startDate->modify('+1 day');
+
+        $endDate = clone $startDate;
```
slizzio: same here. if its just the dates we need then I dont think you need to call clone

tests/AppBundle/Entity/Report/ReportTest.php
```diff
@@ -27,15 +28,48 @@ class ReportTest extends \PHPUnit_Framework_TestCase
...
+        $this->validReportCtorArgs = [$this->client, Report::TYPE_102, new \DateTime('2017-06-23'), new \DateTime('2018-06-22')];
+        $this->report = m::mock(Report::class . '[has106Flag]', $this->validReportCtorArgs);
 
         $this->gift1 = m::mock(Gift::class, ['getAmount' => 1]);
         $this->gift2 = m::mock(Gift::class, ['getAmount' => 10]);
         $this->expense1 = m::mock(Expense::class, ['getAmount' => 2]);
         $this->expense2 = m::mock(Expense::class, ['getAmount' => 20]);
     }
 
+    public static function ctorProvider()
```
slizzio: Q: whats ctor? (probably something obvious :-)

tests/AppBundle/Entity/Report/ReportTest.php
```diff
@@ -27,15 +28,48 @@ class ReportTest extends \PHPUnit_Framework_TestCase
...
+    {
+        return [
+            // start date, end date, submitted (true/false)
+            ['2017-06-23', '2018-06-22', [['2016-06-23', '2017-06-22', false]], 'already has unsubmitted report'],
+            ['2017-06-23', '2018-06-24', [['2016-06-23', '2017-06-22', true]], 'cannot cover more than one year'],
+            ['2017-06-24', '2018-06-23', [['2016-06-23', '2017-06-22', true], ['2015-06-23', '2016-06-22', true]], 'new report is expected to start on'],
```
slizzio: any chance of a few more tests here... passing leap dates etc. There is a lot of logic in the constructor and a few more date checks may prevent future maintenance?

tests/AppBundle/Entity/Report/ReportTest.php
```diff
@@ -27,15 +28,48 @@ class ReportTest extends \PHPUnit_Framework_TestCase
...
+        }
+
+        $exceptionMessage = '-';
+        try {
+            new Report($client, Report::TYPE_102, new \DateTime($startDate), new \DateTime($endDate));
+        } catch(\Exception $e){
+            $exceptionMessage = $e->getMessage();
+        }
+
+        $this->assertContains($expectedTextInException, $exceptionMessage);
```
slizzio: Just for info, you can use $this->setExpectedException($exceptionMessage);

src/AppBundle/Service/ReportService.php
elvisciotti: it needs a clone, as modify acts on the object, and if $oldReport is flushed by mistake, the old report date will be altered

tests/AppBundle/Entity/Report/ReportTest.php
elvisciotti: ctor=constructor. I'll rename it

tests/AppBundle/Entity/Report/ReportTest.php
elvisciotti: as per slack channel, too complicated to do on this PR. And not even sure it's worth it atm. For PA we create those dates so there is no risk of failure, and Lay deputies would need a clear message if interval is wrong (arleady done by the FE)

tests/AppBundle/Entity/Report/ReportTest.php
elvisciotti: ah nice, never used it. Changed it

src/AppBundle/Controller/Report/ReportController.php
elvisciotti: Problem is that that method is also used for creating new year reports (PA included), where the default value (happening for PA that don't have a CASREC entry at the submit time) is the old year report's type. Report type will at some point require that casrec entry, and at that time we can refactor this without using a default value

slizzio: ok I thought that might be the case

PR #337, opened 2017-07-06
Ddpb 814 upload

slizzio:


tests/AppBundle/Controller/DocumentControllerTest.php
```diff
@@ -0,0 +1,66 @@
...
+        // assert POST for deputy
+        $data = $this->assertJsonRequest('POST', $url, [
+            'mustSucceed' => true,
+            'AuthToken'   => self::$tokenDeputy,
+            'data'        => [
+                'file_name'=> 'testfile.pdf',
+                'storage_reference'   => 's3StorageKey'
+            ],
+        ])['data'];
+    }
```
elvisciotti: @slizzio we could add here some assertions on the added document. e.g. take it from doctrine by ID, and assert the two properties are set

slizzio: Yes I was thinking of that at the time. I'll add it

slizzio: Added full assertion list, had to remove the entity reference to repository that doesnt exist. Should we have these? I can create an empty Repository class and leave the reference in the entity and test will pass

PR #340, opened 2017-07-20
Ddpb 815 remove documents

slizzio:


src/AppBundle/Controller/Report/DocumentController.php
```diff
@@ -37,4 +34,57 @@ public function add(Request $request, $reportId)
...
+            $document = $this->findEntityBy(Document::class, $id);
+
+            // enable if the check above is removed and the note is available for editing for the whole team
+            $this->denyAccessIfClientDoesNotBelongToUser($document->getReport()->getClient());
+
+            $this->getEntityManager()->remove($document);
+
+            $this->getEntityManager()->flush($document);
+        } catch (\Exception $e) {
+            $this->get('logger')->error('Failed to delete document ID: ' . $id . ' - ' . $e->getMessage());
```
elvisciotti: exception should be re-thrown in order for the listener to convert it into a readable response for the client

PR #342, opened 2017-07-24
Ddpb 1395 - download area + refactors

elvisciotti:


app/DoctrineMigrations/Version139.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE TABLE report_submission (id SERIAL NOT NULL, report_id INT DEFAULT NULL, archived_by INT DEFAULT NULL, created_by INT DEFAULT NULL, archived BOOLEAN DEFAULT \'false\' NOT NULL, created_on TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL, PRIMARY KEY(id))');
```
slizzio: Do we really need archived BOOLEAN DEFAULT \'false\'  ? The archived by will indicate this, if set then archived. If not set then not archived?

app/DoctrineMigrations/Version139.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->addSql('CREATE INDEX IDX_C84776C84BD2A4C0 ON report_submission (report_id)');
+        $this->addSql('CREATE INDEX IDX_C84776C851B07D6D ON report_submission (archived_by)');
+        $this->addSql('CREATE INDEX IDX_C84776C8DE12AB56 ON report_submission (created_by)');
+        $this->addSql('ALTER TABLE report_submission ADD CONSTRAINT FK_C84776C84BD2A4C0 FOREIGN KEY (report_id) REFERENCES report (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
```
slizzio: Is there any chance a report submission will get removed and remove the report? with 'REFERENCES report (id) ON DELETE CASCADE'

app/config/security.yml
```diff
@@ -86,6 +86,7 @@ security:
             stateless:   true
 
     role_hierarchy:
+        ROLE_ADMIN:           ROLE_DOCUMENT_UPLOAD
```
slizzio: Just a thought, but wouldnt ROLE_DOCUMENT_DOWNLOAD be a better choice for admins? Since they can only download. All deputies have the permission to UPLOAD. Could get confusing

src/AppBundle/Entity/Report/Document.php
```diff
@@ -140,13 +156,34 @@ public function getReport()
     }
 
     /**
-     * @param Report $report
+     * @param Report|null $report
      *
      * @return $this
      */
-    public function setReport(Report $report)
+    public function setReport(Report $report = null)
```
slizzio: Q: Why would you ever have a document not attached to a report?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -217,6 +219,12 @@ class Report
     private $documents;
 
     /**
+     * @JMS\Type("array<AppBundle\Entity\Report\ReportSubmission>")
+     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Report\ReportSubmission", mappedBy="report")
+     */
+    private $submissions;
```
slizzio: reportSubmissions?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -712,21 +721,41 @@ public function getStatus()
...
      */
-    public function setDocuments($documents)
+    public function addDocument(Document $document)
     {
-        $this->documents = $documents;
-        return $this;
+        if (!$this->documents->contains($document)) {
+            $this->documents->add($document);
+        }
     }
```
slizzio: Ideally addDocument should be a separate method to setDocuments(). Doesn't JMS / Doctrine reference getters and setters in order to hydrate?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -712,21 +721,41 @@ public function getStatus()
...
+
+    /**
+     * @return mixed
+     */
+    public function getSubmissions()
+    {
+        return $this->submissions;
```
slizzio: reportSubmissions?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -712,21 +721,41 @@ public function getStatus()
...
+    }
+
+    /**
+     * @param ReportSubmission $submissions
+     * @return Report
+     */
+    public function addSubmissions(ReportSubmission $submission)
+    {
+        if (!$this->submissions->contains($submission)) {
+            $this->submissions->add($submission);
```
slizzio: again here reportSubmissions

src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -0,0 +1,188 @@
...
+    public function getReport()
+    {
+        return $this->report;
+    }
+
+    /**
+     * @param Report $report
+     * @return ReportSubmission
+     */
+    public function setReport($report)
```
slizzio: setReport(Report $report)

src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -0,0 +1,188 @@
...
+    public function getArchivedBy()
+    {
+        return $this->archivedBy;
+    }
+
+    /**
+     * @param User $archivedBy
+     * @return ReportSubmission
+     */
+    public function setArchivedBy($archivedBy)
```
slizzio: setArchivedBy(User $archivedBy)

src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -0,0 +1,188 @@
...
+        return $this;
+    }
+
+
+    /**
+     * @return boolean
+     */
+    public function isArchived()
+    {
+        return $this->archived;
```
slizzio: You can lose the archived column as its unnecessary. 
return ($this->archiveBy instanceof User) should do it;

src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * @param boolean $archived
+     */
+    public function setArchived($archived)
+    {
+        $this->archived = $archived;
```
slizzio: return $this;

src/AppBundle/Service/ReportService.php
```diff
@@ -109,7 +111,6 @@ public function createNextYearReport(Report $oldReport)
         }
 
         $this->_em->persist($newReport);
-        $this->_em->flush();
```
slizzio: removing this will not persist the newReport unless a flush is called elsewhere. Risky.

slizzio: Forget this, I see now why youve done it.

tests/AppBundle/Service/ReportServiceTest.php
```diff
@@ -5,31 +5,34 @@
...
 
-class ReportServiceTest extends m\Adapter\Phpunit\MockeryTestCase
+class ReportServiceTest extends \PHPUnit_Framework_TestCase
 {
     /**
-     * @var Client
+     * @var Report
      */
-    protected $frameworkBundleClient;
+    private $user;
```
slizzio: $user or Report?

src/AppBundle/Entity/Report/Report.php
elvisciotti: yeah, I'll rename that

src/AppBundle/Entity/Report/Report.php
elvisciotti: I left it on purpose to keep it simple, as not needed. Doctrine already creates the setters in the proxies when missing. JMS doesn't need them as the API as itonly reads those documents, and never sets them in block (unless when created manually). JMS for setter in general is not required in the API as we set them manually in the endpoint.
I can still add the setters, or maybe - if we want it for all the entities - we create a ticket to go through all the entities ?

src/AppBundle/Entity/Report/Report.php
elvisciotti: changed

src/AppBundle/Entity/Report/ReportSubmission.php
elvisciotti: changed

src/AppBundle/Entity/Report/ReportSubmission.php
elvisciotti: this one is initially null, I'll add
public function setArchivedBy(User $archivedBy = null)

src/AppBundle/Entity/Report/ReportSubmission.php
elvisciotti: yeah I was thinking the same. I'll drop it. should be easy in the endpoint to filter the query. I'll add a test for report-submission endpoint too

src/AppBundle/Entity/Report/ReportSubmission.php
elvisciotti: method will be removed

src/AppBundle/Service/ReportService.php
elvisciotti: yeah, single transaction, better to fail than having the db dirty. Private method, so we have visibility on the calls

tests/AppBundle/Service/ReportServiceTest.php
elvisciotti: USer. cahnged

app/DoctrineMigrations/Version139.php
elvisciotti: that property will go

app/DoctrineMigrations/Version139.php
elvisciotti: that should mean that the removed report will remove the submissions. I don't think report are ever deleted though

slizzio: You could well be right about the cascade which would be desired behaviour by deleting a report. But I thought the cascade needed to go on the report.report_submission side to get what you want... like i said you may well be right... just seems odd to specify it on the inverse side I cant imagine a report submission ever being deleted  either though. Ide be tempted to leave records untouched until we have a use case for deleting either. You cant delete a submission anyway at the moment can you?

src/AppBundle/Entity/Report/Document.php
slizzio: connecting both gives an easier way of seeing ALL the documents attached to the report, regardless of when they were submitted.
So to get all documents you'd do $report->getDocuments() and to get what was attached when you'd loop through report->getReportSubmissions() calling getDocuments on each submission It shouldnt be a repition because each document stores both the report and submission id, its two different 1:n relationships. Id leave the documents attached to both and by removing the null we also ensure enforce the fact that each document MUST be attached to a report (hence its call in the constructor)

src/AppBundle/Entity/Report/Report.php
slizzio: No thats fine. I don't believe in creating methods just for the sake of it. Hence my question whether JMS relies on them (I think Zend might)

PR #361, opened 2017-09-04
DAPH-177 consume latest docker image

Azulinho:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.service.opg.digital/opguk/php-fpm:0.1.216
+FROM registry.service.opg.digital/opguk/php-fpm
```
elvisciotti: @Azulinho what are the changes?

PR #368, opened 2017-09-15
Ddpb 1053 joint report support

elvisciotti:


src/AppBundle/Controller/BehatController.php
```diff
@@ -23,6 +26,26 @@ private function securityChecks()
     }
 
     /**
+     * @Route("/client/{caseNumber}")
+     * @Method({"PUT"})
+     */
+    public function clientEditAction(Request $request, $caseNumber)
```
slizzio: Does this need additional security checks?

src/AppBundle/Entity/CasRec.php
```diff
@@ -227,20 +227,37 @@ public function getCorref()
...
+
+        if (in_array($userRoleName, [User::ROLE_PA, User::ROLE_PA_TEAM_MEMBER, User::ROLE_PA_ADMIN])) {
+            if (in_array($corref, ['l3', 'l3g', 'a3']) && $typeOfRep === 'opg103') {
+                return Report::TYPE_103_6;
+            }
+            if (Report::ENABLE_104 && $corref === 'hw' && $typeOfRep === '') {
+                return Report::TYPE_104_6;
+            }
+
+            return Report::TYPE_102_6; //default for PA
```
slizzio: Id be tempted to throw an exception if the type cannot be determined by logic. Rather than have incorrect report types in the database where we wouldnt neccessarily be aware

src/AppBundle/Entity/CasRec.php
```diff
@@ -227,20 +227,37 @@ public function getCorref()
...
         }
 
-        return Report::TYPE_102;
+        throw new \Exception(__METHOD__ . ": user role not recognise to determine report type");
```
slizzio: Like this :) only the return statement will prevent this from being executed

src/AppBundle/Controller/BehatController.php
elvisciotti: it's done by `securityChecks` below. Basically completely disabled on non-prod envs

src/AppBundle/Entity/CasRec.php
elvisciotti: there still is a default to 102 for each user role, but we can think of having an exception once the CSV fields are more clear

PR #370, opened 2017-09-21
Co deputies

jackgoodby:


src/AppBundle/Controller/CoDeputyController.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->denyAccessUnlessGranted([EntityDir\User::ROLE_DEPUTY]);
+
+        $data = $this->deserializeBodyContent($request, [
+            'email' => 'notEmpty',
+        ]);
+
+        $loggedInUser = $this->getUser();
+        $newUser = new EntityDir\User();
+
+        $newUser->setFirstname('');
```
elvisciotti: the logic from here could go in a method of the userService

src/AppBundle/Controller/CoDeputyController.php
```diff
@@ -0,0 +1,80 @@
...
+        $newUser->setLastname('');
```
slizzio: wont a new User object have an empty first and last name? Not sure above 2 statements are required

src/AppBundle/Controller/SelfRegisterController.php
```diff
@@ -48,6 +49,42 @@ public function register(Request $request)
...
+            $coDeputyVerified = $this->container->get('user_registration_service')->validateCoDeputy($selfRegisterData);
+            $this->get('logger')->warning('CasRec codeputy validation success: ', ['extra' => ['page' => 'codep_validation', 'success' => true] + $selfRegisterData->toArray()]);
+        } catch (\Exception $e) {
+            $this->get('logger')->warning('CasRec codeputy validation failed:', ['extra' => ['page' => 'codep_validation', 'success' => false] + $selfRegisterData->toArray()]);
+throw $e;
+        }
+
+return print_r(doctrineDebug::export($coDeputyVerified, 3),true);
+
+        return ['verified' => $coDeputyVerified];
```
slizzio: above 6 lines need a reformat. remove last return statement

src/AppBundle/Entity/User.php
```diff
@@ -986,6 +995,44 @@ public function getAgreeTermsUseDate()
...
+     * @return bool
+     */
+    public function isCoDeputy()
+    {
+        $isCoDeputy = false;
+        if ($this->isLayDeputy()) {
+            $client = $this->getFirstClient();
+            if (!empty($client)) {
+                $isCoDeputy = count($client->getUsers()) > 1;
+            }
```
slizzio: Maybe throw an exception if client not set?

src/AppBundle/Controller/SelfRegisterController.php
jackgoodby: that was debug! good spot

PR #372, opened 2017-10-02
Ddpb 1323 upload docs submitted report

slizzio:


src/AppBundle/Service/ReportService.php
```diff
@@ -203,4 +203,31 @@ public function submit(Report $currentReport, User $user, \DateTime $submitDate)
...
+    /**
+     * Set report submission for additional documents
+     *
+     * @param Report $currentReport
+     * @param User $user
+     * @param \DateTime $submitDate
+     *
+     * @return Report new year's report
+     */
+    public function submitAdditionalDocuments(Report $currentReport, User $user, \DateTime $submitDate)
```
elvisciotti: a unit test should be useful here. ReportServiceTest is there already

slizzio: Added Unit test

PR #373, opened 2017-10-06
Ddpb 1645

elvisciotti:


app/DoctrineMigrations/Version153.php
```diff
@@ -0,0 +1,36 @@
...
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE account DROP COLUMN IF EXISTS opening_date');
+        $this->addSql('ALTER TABLE account DROP COLUMN IF EXISTS closing_date');
+        $this->addSql('ALTER TABLE account DROP COLUMN IF EXISTS opening_date_explanation');
+        $this->addSql('ALTER TABLE account DROP COLUMN IF EXISTS closing_balance_explanation');
+        $this->addSql('ALTER TABLE account DROP COLUMN IF EXISTS closing_date_explanation');
```
slizzio: Is there a chance we'll lose some data with this?

src/AppBundle/Entity/Client.php
```diff
@@ -44,14 +44,14 @@ class Client
      *
      * @JMS\Groups({"client-reports"})
      * @JMS\Type("array")
-     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Report\Report", mappedBy="client", cascade={"persist"})
+     * @ORM\OneToMany(targetEntity="AppBundle\Entity\Report\Report", mappedBy="client", cascade={"persist", "remove"})
```
slizzio: Dont we had a check for this within admin. Are we sure we want to do this? :)

src/AppBundle/Service/ReportService.php
```diff
@@ -203,4 +205,44 @@ public function submit(Report $currentReport, User $user, \DateTime $submitDate)
...
+        }
+
+        $ret = [];
+        foreach($reports as $report1) {
+            foreach($reports as $report2) {
+                if ($report1->getId() === $report2->getId()) {
+                    continue;
+                }
+                // find report with same date that have not started
+                if ($report1->getStatus()->hasStarted()
```
slizzio: Should this be negated? comment says    // find report with same date that have not started

src/AppBundle/Service/ReportStatusService.php
```diff
@@ -547,6 +547,22 @@ public function getSubmitState()
...
+     */
+    public function hasStarted()
+    {
+        $sectionStatus = $this->getSectionStatus();
+        // exclude balance, and money transfers, that depend on other section therefore not required
+        // to complete and therefore considered "done"
+        unset($sectionStatus['balance']);
+        unset($sectionStatus['moneyTransfers']);
+
+        return count(array_filter($sectionStatus, function ($e) {
```
slizzio: typecast the count to (bool)

src/AppBundle/Entity/Client.php
elvisciotti: there is a db cascade delete  on client.report, and this is corresponding application-level cascade (only useful to keep the unity of work clean, as the cascade delete logic is already applied by the db). I needed this cascade to delete clients without users

app/DoctrineMigrations/Version153.php
elvisciotti: columns have been removed lots of time ago, as the UX doesn't require nor used this data anymore. Very unlikely we need that, as old reports have been already submitted and it's not required for new reports

PR #376, opened 2017-10-10
Ddpb 1511 codeps selfregister

jackgoodby:


src/AppBundle/Service/UserRegistrationService.php
```diff
@@ -25,39 +25,49 @@ public function __construct($em)
...
-        if (!$this->userIsUnique($user)) {
-            throw new \RuntimeException("User with email {$user->getEmail()} already exists.", 422);
+        // Check the user doesn't already exist
+        $existingUser = $this->em->getRepository('AppBundle\Entity\User')->findOneBy(['email' => $selfRegisterData->getEmail()]);
+        if ($existingUser) {
+            throw new \RuntimeException("User with email {$existingUser->getEmail()} already exists.", 422);
         }
 
         // Check the client is unique
-        if (!$this->clientIsUnique($client)) {
```
slizzio: Should remove the comment too if unique check is removed

jackgoodby: no - the check still takes place at that point - just refactored

PR #385, opened 2017-10-26
Ddpb 1680 imporoved reg validation

slizzio:


src/AppBundle/Controller/CasRecController.php
```diff
@@ -110,6 +111,29 @@ public function addBulk(Request $request)
         return ['added' => $added - 1, 'errors' => $retErrors];
     }
 
+
+    /**
+     * Verify Deputy & Client last names, Postcode, and Case Number
+     *
+     * @Route("/verify")
+     * @Method({"POST"})
+     */
```
slizzio: It would be useful to have the @return value here

src/AppBundle/Service/CasrecVerificationService.php
```diff
@@ -0,0 +1,159 @@
...
+    {
+        $crMatches = $this->casRecRepo->findBy( [ 'caseNumber'     => $this->normaliseCaseNumber($caseNumber)
+                                                , 'clientLastname' => $this->normaliseSurname($clientSurname)
+                                                , 'deputySurname'  => $this->normaliseSurname($deputySurname)
+                                                ]);
+
+        $this->lastMatchedCasrecUsers = $this->applyPostcodeFilter($crMatches, $deputyPostcode);
+
+        if (count($this->lastMatchedCasrecUsers) == 0) {
+            throw new \RuntimeException('User registration: no matching record in casrec', 400);
```
slizzio: im happy to leave this for now given the size of the story but an exception is something we dont expect to happen. Given the validate nature of this method simply returning false would work better here. Since the method succeeded to run and simply didnt validate

src/AppBundle/Service/CasrecVerificationService.php
```diff
@@ -0,0 +1,159 @@
...
+
+        return true;
+    }
+
+    /**
+     * Since co-deputies, multiple deputies may be matched (eg siblings at same postcode)
+     *
+     * @return array
+     */
+    public function getLastMatchedDeputyNumbers()
```
slizzio: should this be public? how does lastMatchedCasrecUsers get set?

tests/AppBundle/Service/CasrecVerificationServiceTest.php
```diff
@@ -0,0 +1,201 @@
...
+            $this->assertTrue($this->casrecVerificationService->validate('11111111', 'CSurn', 'WRONG', 'DPC123'));
+        } catch(\RuntimeException $e) {
+            $this->assertEquals($failMessage, $e->getMessage());
+        }
+
+        try {
+            $this->assertTrue($this->casrecVerificationService->validate('11111111', 'CSurn', 'DSurn', 'WRONG'));
+        } catch(\RuntimeException $e) {
+            $this->assertEquals($failMessage, $e->getMessage());
+        }
```
slizzio: Better to have these as separate tests. One for sucess, one for failures and one testing for exceptions... Pretty good for a first attempt though... instead of using try catch you can use $this->setExpectedException()

src/AppBundle/Service/UserRegistrationService.php
```diff
@@ -7,20 +7,19 @@
 use AppBundle\Entity\User;
 use AppBundle\Model\SelfRegisterData;
 use Doctrine\ORM\EntityManager;
-use Doctrine\Common\Util\Debug as doctrineDebug;
+use \Doctrine\Common\Util\Debug as doctrineDebug;
```
elvisciotti: let me know when merged, I'll run php-cs-fixer to remove unused namespaces

PR #388, opened 2017-11-20
DDPB-1760-stats-add-column casrec.lastReportSubmittedAt + CSV writing…

elvisciotti:
… fixes

### DIFF
--- a/src/AppBundle/Service/CasrecService.php
+++ b/src/AppBundle/Service/CasrecService.php
@@ -49,7 +49,12 @@ class CasrecService
         $this->validator = $validator;
     }
 
-
+    /**
+     * Updates a single CASREC record, with stats
+     * Called when records are uploaded, or "updateAllCasrecRecordsWithStats" is called via cron
+     *
+     * @param CasRec $casrec
+     */
     private function updateCasrecStatsSingle(CasRec $casrec)
     {
         // add user info, by matching DeputyNo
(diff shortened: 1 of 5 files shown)


src/AppBundle/Service/CasrecService.php
```diff
@@ -69,8 +74,14 @@ private function updateCasrecStatsSingle(CasRec $casrec)
             ->setParameter('c2', strtolower($caseNumber))
             ->getResult();
         if ($results && $results[0] instanceof Client) {
+            $client = $results[0]; /* @var $client Client */
+            // last report is currently ordered
+            $submittedReports = $client->getSubmittedReports();
+            $lastReport = count($submittedReports) > 0  ? $submittedReports->first() : null;
```
slizzio: Optional: If report order changes for any reason (getSubmittedReports()) this will screw up the stats. I know its extra work here but is it worth reordering here to ensure the order is still latest first here?

elvisciotti: getSubmittedReports orders them already. I was thinking about changing the name of that method. It's misleading as getSubmittedReports should just return the reports, If it returns them ordered, than it should be called differently

slizzio: maybe a new method getLatestSubmittedReports(). The change was optional, im just conscious that if any other dev picks up a story to change the ordering (to last updated for example), then they wont be aware of this dependency

PR #391, opened 2017-11-24
Ddpb 1626 deny admin password reset

slizzio:


src/AppBundle/Controller/UserController.php
```diff
@@ -299,6 +299,10 @@ public function recreateToken(Request $request, $email, $type)
         }
         $user = $this->findEntityBy(EntityDir\User::class, ['email' => $email]);
 
+        if (!$this->getAuthService()->isSecretValidForRole($user->getRoleName(), $request)) {
+            throw new \RuntimeException('Admin emails not accepted.', 403);
+        }
+
```
slizzio: could this role check replace L297-300?

tests/AppBundle/Controller/UserControllerTest.php
```diff
@@ -401,6 +424,24 @@ public function testRecreateTokenUserNotFound($urlPart)
         ]);
     }
 
+    /**
+     * @dataProvider recreateTokenProviderForRole
+     */
+    public function testRecreateTokenUserIsAdmin($urlPart, $secret, $email, $passOrFail)
```
elvisciotti: why just just $passOrFail ? mustSucceed' : 'mustFail'
as the key ? less code

PR #400, opened 2017-12-22
Ddpb 1839 new debts question

slizzio:


src/AppBundle/Entity/Report/Traits/DebtTrait.php
```diff
@@ -47,6 +57,22 @@ public function getDebts()
...
+    /**
+     * @return mixed
+     */
+    public function getDebtManagement()
+    {
+        return $this->debtManagement;
+    }
+
+    /**
+     * @param mixed $debtManagement
```
elvisciotti: string

src/AppBundle/Entity/Odr/Odr.php
```diff
@@ -331,6 +340,22 @@ public function addDebt(Debt $debt)
         return $this;
     }
 
+    /**
+     * @return mixed
```
elvisciotti: string

app/DoctrineMigrations/Version159.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
elvisciotti: we don't use down migration, so we can remove all 3 lines in the future. not needed if merged already. I'd not work anyway due to the first

PR #402, opened 2017-12-27
DDPB 1762 - NDR submitted as ZIP

elvisciotti:


src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -78,13 +91,19 @@ class ReportSubmission
...
-        $this->report = $report;
-        $this->report->addReportSubmission($this);// double-link for UNIT test purposes
+        if ($report instanceof Report) {
+            $this->report = $report;
+            $this->report->addReportSubmission($this);// double-link for UNIT test purposes
+        } else if ($report instanceof Odr) {
+            $this->ndr = $report;
+        } else {
+            throw new \InvalidArgumentException(__METHOD__.' first argumnt should be a Report or an Ndr');
+        }
```
slizzio: Nothing to change really as its only a minor comment but using a ReportInterface instead of an AbstractReport in the constructor signature would mean we can get rid of all this checking and replace the signature with: 
__construct(ReportInterface $report, User $createdBy).

elvisciotti: changed AbstractReport into ReportInterface

PR #409, opened 2018-01-18
Ddpb 1838 client archive

jackgoodby:


src/AppBundle/Controller/ClientController.php
```diff
@@ -6,6 +6,7 @@
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Component\HttpFoundation\Request;
+use \Doctrine\Common\Util\Debug as doctrineDebug;
```
elvisciotti: to remove

jackgoodby: Cheers - done

PR #416, opened 2018-02-06
Ddpb 1768 search clients

slizzio:


src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -0,0 +1,53 @@
...
+
+        $qb = $this->createQueryBuilder('c');
+        $qb->setFirstResult($offset);
+        $qb->setMaxResults($limit);
+        $qb->leftJoin('c.users', 'u')
+            ->where('u.roleName = \'' . User::ROLE_LAY_DEPUTY . '\'');
+        $qb->orderBy('c.' . $orderBy, $sortOrder);
+
+        if ($query) {
+            if (preg_match('/^[0-9t]{8}$/i', $query)) { // case number
```
elvisciotti: minor: this is useful and could be moved to a static method into the Client class, and re-used. Might be used already around

src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @param string $query Search query
+     * @param string $orderBy field to order by
+     * @param string $sortOrder order of field order ASC|DESC
+     * @param int $limit number of results to return
+     * @param string $offset
+     *
+     * @return Client[]|array
+     */
+    public function searchClients($query = '', $orderBy = 'lastname', $sortOrder = 'ASC', $limit = 100, $offset = 'id')
```
elvisciotti: default param "id", should be 0

elvisciotti: since it's called with all the params, I don't think it's bad to remove default params. we'll shift the responsability to the caller, and simplify things

slizzio: Yes, copied from somewhere else. Ill change. Need to force the limit to be max 100 too

src/AppBundle/Entity/Repository/ClientRepository.php
slizzio: it IS used for the user search, but one issue we've got is that none of our test data matches it lol. as they all have case numbers with the word "test" so no room for numbers

slizzio: Moved to Client::isValidCaseNumber($q)

PR #419, opened 2018-02-12
Ddpb 857 unsubmit reports

elvisciotti:


src/AppBundle/Entity/Report/Report.php
```diff
@@ -245,6 +245,15 @@ public static function getSectionsSettings()
...
     private $submitDate;
 
+    /**
+     * @var \DateTime
+     *
+     * @JMS\Groups({"report"})
+     * @JMS\Type("DateTime")
+     * @ORM\Column(name="un_submit_date", type="datetime", nullable=true)
+     */
+    private $unSubmitDate;
```
slizzio: Is the unsubmitDate used anywhere? At the moment we have 2 flags to indicate whether a report is submitted, the submit_date and submitted flag. Unless its required to store this field the same can be achieved by setting the submit_date to null and submitted to false. My concern is where we need to identify reports that have been submitted we'll need to check 3 fields instead of 1.

elvisciotti: atm the UI requires to differentiate between report never been submitted, and report being submitted at least once. We'll get back to refactor the other fields when it's all done. Report.submit itself could be removed as submitDate being null or not already tell if submitted or not

elvisciotti: when unsubmitting the report.submit will be set to false, so no need to change any current filters/queries anywhere. We'll review the whole thing when we actually implement the flagged being stored

PR #438, opened 2018-04-09
Ddpb 2045 account dropdown

slizzio:


src/AppBundle/Controller/Report/AccountController.php
```diff
@@ -139,25 +141,99 @@ private function fillAccountData(EntityDir\Report\BankAccount $account, array $d
     }
 
     /**
-     * Check bank account has transfers
+     * Check bank account has transactions
      *
      * @param EntityDir\Report\BankAccount $account
      */
-    protected function denyAccessIfAccountHasTransfers(EntityDir\Report\BankAccount $account)
+    protected function denyAccessIfAccountHasTransactions(EntityDir\Report\BankAccount $account)
```
elvisciotti: is the name still correct considering there are gifts  and transfer checked inside ?

app/DoctrineMigrations/Version173.php
```diff
@@ -0,0 +1,55 @@
...
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE expense ADD bank_account_id INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE expense ADD CONSTRAINT FK_2D3A8DA612CB990C FOREIGN KEY (bank_account_id) REFERENCES account (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('CREATE INDEX IDX_2D3A8DA612CB990C ON expense (bank_account_id)');
+
+        $this->addSql('ALTER TABLE gift ADD bank_account_id INT DEFAULT NULL');
```
elvisciotti: there is "ON DELETE" on  "expense", but not on "gift" table

slizzio: Fixed, forgot to re-run the migrations script after changing the schema. All bank account fields now set to null on deletion of the account. This will tie in with the error messaging next sprint

src/AppBundle/Controller/Report/AccountController.php
slizzio: Yes it was renamed to reflect a "transaction" rather than a specific money changing action. A transaction in this case just means money leaving or going into an account so would cover gifts, transfers, expenses as well as money in and out.

PR #439, opened 2018-04-10
Ddpb 2025 admin notifications

elvisciotti:


tests/AppBundle/Controller/SettingControllerTest.php
```diff
@@ -0,0 +1,119 @@
...
+            self::$tokenDeputy = $this->loginAsDeputy();
+            self::$tokenAdmin = $this->loginAsAdmin();
+        }
+    }
+
+    public function testgetOneByIdNotPresent()
+    {
+        Fixtures::deleteReportsData(['setting']);
+
+        $id = 'service-notification';
```
slizzio: Im a bit confused by id being a string. The entity states its an integer ?

src/AppBundle/Entity/Setting.php
```diff
@@ -0,0 +1,115 @@
...
+use JMS\Serializer\Annotation as JMS;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="setting")
+ */
+class Setting
+{
+    /**
+     * @var int
```
slizzio: string?

src/AppBundle/Entity/Setting.php
```diff
@@ -0,0 +1,115 @@
...
+     * @JMS\Type("string")
+     * @JMS\Groups({"setting"})
+     *
+     * @ORM\Column(name="id", type="string", length=64, nullable=false)
+     * @ORM\Id
+     */
+    private $id;
```
slizzio: I would keep the id as an integer and have a settings_type field which could be a string

elvisciotti: I thought the same, but then I'd need to make it more complicated with the getBy() and find(). ID autoincrement not strictly necessary for a fixed list of settings. What about we review this next time we need to add another setting ? autoincrement Id are not mandatory in doctrine. Ids are, and whether they are strings or int, make no difference in the end

src/AppBundle/Controller/SettingController.php
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param int $id
+     *
+     * @return array
+     */
+    public function delete($id)
+    {
+        try {
+            /** @var $note EntityDir\Note $note */
+            $note = $this->findEntityBy(EntityDir\Note::class, $id);
```
slizzio: note?

src/AppBundle/Entity/Setting.php
slizzio: Im not sure why would it be more complicated? Personally, I find it more complicated to see an id being represented as a string when all other ids are integers. Hence I was confused when i read the tests and cross referenced with the entity doc blocks. Wasnt even sure how it was working at first.
Calls to getBy and find() should simply use :
findOneBy(array('settings_type' => $type);

PR #441, opened 2018-04-23
Ddpb 2058 cross team

elvisciotti:


app/DoctrineMigrations/Version178.php
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE dd_user ADD deleted_at TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT NULL');
```
slizzio: I think this can be removed since it would error if ran. The up just removes records, doesnt remove the field.

elvisciotti: I'll remove it. We don't have a jenkins job to run down migrations. We can either leave them (as autogenerated), or delete

src/AppBundle/Controller/CoDeputyController.php
```diff
@@ -83,8 +83,8 @@ public function update(Request $request, $id)
         if (!empty($data['email'])) {
             $originalUser = clone $user;
             $user->setEmail($data['email']);
-            $userService = $this->get('opg_digideps.user_service');
-            $userService->editUser($originalUser, $user);
+            $userService = $this->get('user_service');
+            $userService->editUser($this->getUser(), $originalUser, $user);
```
slizzio: this->getUser can be accessed within  the service cant it? Parameters are a bit confusing? :)

src/AppBundle/Service/ReportUtils.php
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param \DateTime $reportEndDate
+     *
+     * @return \DateTime $reportStartDate
+     */
+    public static function generateReportStartDateFromEndDate(\DateTime $reportEndDate)
+    {
+        $reportStartDate = clone $reportEndDate;
+
+        $isLeapDay = $reportStartDate->format('d-M') == '29-Feb';
```
slizzio: i think you can use isValidDay on DateTime object for this

slizzio: either that or DateTime returns False if set to invalid day

src/AppBundle/Controller/CoDeputyController.php
elvisciotti: this->getUser is the logged in user . In theory it can be injected ( security.token_storage) into the service, but then the service cannot be constructed if the user is not logged. At this stage I moved thing into the service without changing too much

src/AppBundle/Service/ReportUtils.php
elvisciotti: I've only moved the existing code from a non-static method in the class as it was unrelated the the class (and wasn't even static). I agree it should be refactored, doesn't look a good solution. I'll add a TODO comment to refactor later, branch big enough atm

elvisciotti: That method substract one year to the date, so the return value should always be valid. Not sure about the leapDay thing, needs more investigation and specs collect

PR #446, opened 2018-05-14
Ddpb 2063 account delete conf page

elvisciotti:


src/AppBundle/Controller/Report/AccountController.php
```diff
@@ -77,6 +78,51 @@ public function editAccountAction(Request $request, $id)
...
+                Report::SECTION_MONEY_IN => $report->hasSection(Report::SECTION_MONEY_IN)
+                   ? count($report->getMoneyTransactionsIn()->filter($paymentsFilter))
+                    : null,
+                Report::SECTION_MONEY_OUT => $report->hasSection(Report::SECTION_MONEY_OUT)
+                    ? count($report->getMoneyTransactionsOut()->filter($paymentsFilter))
+                    : null
+            ]
+        ];
+
+        $ret['paymentsTotal'] = array_sum($ret['payments']);
```
slizzio: I may be wrong but doesnt this return the number of transactions involving the bank account? If so then "paymentsTotal" is misleading as this implies the payment totals have been summed, I think transactionCount might be better?

PR #460, opened 2018-07-06
Ddpb 2158 previous report data

Biggsen:


src/AppBundle/Entity/Ndr/BankAccount.php
```diff
@@ -314,6 +315,28 @@ public function getBalanceOnCourtOrderDate()
         return $this->balanceOnCourtOrderDate;
     }
 
+    /**
+     * @return decimal
+     */
+    public function getOpeningBalance() {
```
elvisciotti: we'll run php-cs-fixed for brackets on next line

src/AppBundle/Entity/Report/Report.php
```diff
@@ -1159,4 +1160,102 @@ public function setChecklist($checklist)
...
+            ];
+        }
+
+    /**
+     * Method to identify and return previous report.
+     *
+     * @return Ndr|Report|bool|mixed
+     */
+    private function getPreviousReport() {
+        $clientReports = $this->getClient()->getReports();
```
elvisciotti: alternative could be a query with client=same client, report!=currentTeport and taking the max(id) ,but not sure it's simpler

elvisciotti: or maybe, just iterate client reports excluding the current, index by reportId, then krsort, then take the first (array_shift)

PR #463, opened 2018-07-20
Ddpb 2189 pa checklist

slizzio:


src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -30,8 +30,7 @@ public function searchClients($query = '', $orderBy = 'lastname', $sortOrder = '
         $qb = $this->createQueryBuilder('c');
         $qb->setFirstResult($offset);
         $qb->setMaxResults($limit);
-        $qb->leftJoin('c.users', 'u')
-            ->where('u.roleName = \'' . User::ROLE_LAY_DEPUTY . '\'');
+        $qb->leftJoin('c.users', 'u');
```
elvisciotti: is the user needed ? I don't see any usage of "u"

PR #472, opened 2018-08-17
DDPB-2292 set CourtDate against client based on Last Report Day of CSV

slizzio:


src/AppBundle/Entity/Client.php
```diff
@@ -877,21 +877,26 @@ public function getUnsubmittedReports()
      */
     public function getExpectedReportStartDate()
     {
-        // create new datetime object. Do not alter object courtDate property.
+        // clone datetime object. Do not alter object courtDate property.
         /** @var \DateTime $expectedReportStartDate */
-        $expectedReportStartDate = new \DateTime();
-
         if (!($this->getCourtDate() instanceof \DateTime)) {
```
elvisciotti: maybe this can go in a method getExpectedReportStartDate  ?

slizzio: It is...(?)

PR #475, opened 2018-09-21
DDPB-2310-dashboard-performance

elvisciotti:


src/AppBundle/Entity/Report/Report.php
```diff
@@ -467,6 +478,7 @@ public function __construct(Client $client, $type, \DateTime $startDate, \DateTi
             $statusCached[$sectionId] = ['state' => ReportStatusService::STATE_NOT_STARTED, 'nOfRecords' => 0];
         }
         $this->setStatusCached($statusCached);
+        $this->statusStatusCached = self::STATUS_NOT_STARTED;
```
slizzio: typo? statusStatusCached

src/AppBundle/Service/ReportStatusService.php
```diff
@@ -639,16 +639,17 @@ public function hasStarted()
...
             return 'notStarted';
         }
 
-        if ($this->report->isDue() && $this->isReadyToSubmit()) {
-            return 'readyToSubmit';
+        if ($ignoreIsDue) {
+            return $this->isReadyToSubmit() ? 'readyToSubmit' : 'notFinished';
         } else {
-            return 'notFinished';
+            return $this->report->isDue() && $this->isReadyToSubmit() ? 'readyToSubmit' : 'notFinished';
```
slizzio: constants?

src/AppBundle/Service/ReportStatusService.php
```diff
@@ -639,16 +639,17 @@ public function hasStarted()
...
+    public function getStatus($ignoreIsDue = false)
     {
         if (!$this->hasStarted()) {
             return 'notStarted';
         }
 
-        if ($this->report->isDue() && $this->isReadyToSubmit()) {
-            return 'readyToSubmit';
+        if ($ignoreIsDue) {
+            return $this->isReadyToSubmit() ? 'readyToSubmit' : 'notFinished';
```
slizzio: constants

tests/AppBundle/Controller-Report/ReportControllerTest.php
```diff
@@ -503,7 +503,7 @@ public function testGetAll()
         //assert results
         $this->assertCount(3, $ret['reports']);
         $this->assertEquals('102', $ret['reports'][0]['type']);
-        $this->assertEquals('pa1Client1', $ret['reports'][0]['client']['firstname']);
```
slizzio: why is this taken out?

app/DoctrineMigrations/Version191.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE INDEX report_status_cached_idx ON report (report_status_cached)');
```
slizzio: we already have a report.statusCached. what is the difference between the two?

src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -443,6 +445,37 @@ public function unsubmit(Request $request, $id)
...
+                ->select('r,c,u')
+                ->leftJoin('r.client', 'c')
+                ->leftJoin('c.users', 'u')
+                ->where('u.id = :uid')
+                ->andWhere('r.reportStatusCached IS NULL')
+                ->setParameter('uid', $userId)
+                ->setMaxResults(5)
+                ->getQuery()
+                ->getResult()) && count($reports)
+        ) {
```
slizzio: This line will be executed on every run through of the loop. Needs moving outside the while statement. Do query once, and use results in while loop

slizzio: could you put some comment above this to explain what its updating please?

slizzio: Building queries belongs int the repository class to allow reuse across other controllers

src/AppBundle/Entity/Report/Report.php
```diff
@@ -388,12 +393,18 @@ public static function getSectionsSettings()
     /**
      * @var Checklist
      *
+     * // TODO "report" group is used by deputy side each time a report is loaded. Better to use a new one like report-checklist
+     *
      * @JMS\Groups({"report"})
```
slizzio: I thought this was report-checklist. May have been reverted at some time?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -388,12 +393,18 @@ public static function getSectionsSettings()
...
      * @JMS\Type("AppBundle\Entity\Report\Checklist")
      * @ORM\OneToOne(targetEntity="AppBundle\Entity\Report\Checklist", mappedBy="report", cascade={"persist", "remove"})
      */
     private $checklist;
 
+    const STATUS_NOT_STARTED = 'notStarted';
+    const STATUS_READY_TO_SUBMIT = 'readyToSubmit';
+    const STATUS_NOT_FINISHED = 'notFinished';
+
```
slizzio: these need moving to top of the class

src/AppBundle/Entity/Report/Traits/StatusTrait.php
```diff
@@ -46,28 +60,43 @@ public function setStatusCached(array $status)
...
+     * and also the report.reportStatusCached using the cache
+     *
      * using the `ReportService::getSectionStateNotCached`
      *
      * @param array $sectionIds
      */
     public function updateSectionsStatusCache(array $sectionIds)
     {
         $currentStatus = $this->getStatusCached();
+        $statusObject = $this->getStatus();
```
slizzio: $currentSectionStatus = $this->getStatusCached();
$currentReportStatus = $this->getStatus();

Makes a lot more sense (if Ive understood correctly) :)

src/AppBundle/Entity/Report/Traits/StatusTrait.php
```diff
@@ -46,28 +60,43 @@ public function setStatusCached(array $status)
...
             }
         }
-
         $this->setStatusCached($currentStatus);
+
+        // update report status, using the cached version of the section statuses
+        // Note: the isDue is skipped
+        $this->reportStatusCached = $statusObject
+            ->setUseStatusCache(true)
+            ->getStatusIgnoringDueDate(true);
```
slizzio: Not sure why the due date is affecting the report status?

src/AppBundle/Service/ReportService.php
```diff
@@ -14,25 +15,48 @@
...
 class ReportService
 {
     /** @var EntityRepository */
     protected $reportRepository;
 
     /**
-     * @var CasRecRepository
+     * @var EntityManager
+     */
+    protected $_em;
```
slizzio: $em is already passed to the constructor

src/AppBundle/Service/ReportService.php
```diff
@@ -296,4 +324,84 @@ public function findDeleteableReports(Collection $reports)
...
+     * @param string $status see Report::STATUS_* constants
+     * @param integer $userId
+     * @param boolean $exclude_submitted
+     * @param string $q search query client firstname/lastname or case number
+     *
+     * @return QueryBuilder
+     */
+    public function getAllReportsQb($select, $status, $userId, $exclude_submitted, $q)
+    {
+        $qb = $this->_em->getRepository(Report::class)
```
slizzio: All repository classes have access to the entity manager. And the reportRepository is already constructed. Hence this method should be in the repository, along with anything else that constructs a query.

src/AppBundle/Service/ReportService.php
```diff
@@ -296,4 +324,84 @@ public function findDeleteableReports(Collection $reports)
...
+        } else if ($status == Report::STATUS_NOT_STARTED) {
+            $qb->andWhere('r.reportStatusCached = :status')
+                ->setParameter('status', $status);
+        }
+
+        return $qb;
+    }
+
+    /**
+     * If the report is ready to submit, but is not yet due, return notFinished instead
```
slizzio: I can see what you're trying to avoid here. But this may lead to lots of people asking why they cant file a report early. Could we have a notDueYet status perhaps?

src/AppBundle/Service/ReportService.php
```diff
@@ -296,4 +324,84 @@ public function findDeleteableReports(Collection $reports)
...
+     * In all the the cases, return original $status
+     *
+     * @param $status
+     * @param \DateTime $endDate
+     *
+     * @return string
+     */
+    public function adjustReportStatus($status, \DateTime $endDate)
```
slizzio: This method shouldnt be needed if reportStatus object is doing its job properly?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -878,20 +890,19 @@ public function setAgreedBehalfDeputyExplanation($agreedBehalfDeputyExplanation)
         return $this;
     }
 
-    public function isDue()
+    /**
+     * @param \DateTime|null $endDate
+     * @return bool
+     */
+    public static function isDue(\DateTime $endDate = null)
```
slizzio: why static? A report must have an end date from when its created. The only thing that end date is compared against is the current date which we will always know.

app/DoctrineMigrations/Version191.php
elvisciotti: report.statusCached holds an array with section statuses, the first solution I implemented, that turned out not to be enough. I had to store the final report status  "report_status_cached" and use an index on them in order to avoid fetching all the reports to calculate the ones to display in the page, and the counts for the tabs

src/AppBundle/Controller/Report/ReportController.php
elvisciotti: I need to execute at each loop, as each loop flushes 5 records (and help at next reload in case of probable crash at first load). This code can be removed when all the reports have a cache, so I didn't put in a repo for reuse

src/AppBundle/Entity/Report/Report.php
elvisciotti: don't know. It probably contributed to slow down the dashboard as it loaded lots of data, but difficult to say, the object hydration was taking too much data anyway. Not a problem anymore with the current solution, the dashboard was teh only point where we load more than 2 reports, and it now just runs 4 simple queries

src/AppBundle/Entity/Report/Report.php
elvisciotti: ah yes, just moved and pushed

src/AppBundle/Entity/Report/Report.php
elvisciotti: yes. fixed and pushed. PHPStorm didn't notice, weird

src/AppBundle/Entity/Report/Report.php
elvisciotti: I need to reuse the same "midnight" logic from the report after the array hydration. It just depends on the end date and related to report, needed in services not connected, so I placed it there. better solution in mind ?

src/AppBundle/Entity/Report/Report.php
```diff
@@ -1267,4 +1278,7 @@ public function getReportTitle()
 
         return $titleTranslationKeys[$this->getType()];
     }
+
+
+
```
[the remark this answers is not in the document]
elvisciotti: removed

src/AppBundle/Entity/Report/Traits/StatusTrait.php
```diff
@@ -26,6 +26,20 @@
      */
     private $statusCached;
```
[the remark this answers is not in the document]
elvisciotti: yes it should be renamed. I had to limit the amount of refactors or this PR will never end :D.
I was actually considering removing the section status. Since we now store the final status, the section status is basically useless. It can be fully calculated in a few ms from the API at each report edit operation with a small overhead and much less code

src/AppBundle/Entity/Report/Traits/StatusTrait.php
```diff
@@ -26,6 +26,20 @@
...
+     * Holds a copy of result of the ReportStatusService::getStatus() results
+     * Used for ORG dashboard for tab calculation and pagination
+     *
+     * value: STATUS_* constant
+     *
+     * @JMS\Exclude()
+     *
+     * @ORM\Column(name="report_status_cached", type="string", length=20, nullable=true)
+     */
+    private $reportStatusCached;
```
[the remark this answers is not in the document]
elvisciotti: not sure I understood. reportStatusCached holds the report status string, yes

src/AppBundle/Entity/Report/Traits/StatusTrait.php
```diff
@@ -46,28 +60,43 @@ public function setStatusCached(array $status)
...
     }
 
+    /**
+     * @return string
+     */
+    public function getReportStatusCached()
+    {
+        return $this->reportStatusCached;
+    }
+
```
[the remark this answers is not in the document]
elvisciotti: simple getter of the reportStatusCached, discussed above. It's not even used as hydrates  as array, but I'll keep it just in case I forgot something and avoid re-running tests, flagged for refactor at next iteration

src/AppBundle/Entity/Report/Traits/StatusTrait.php
elvisciotti: yes, good idea. renamed

src/AppBundle/Entity/Report/Traits/StatusTrait.php
elvisciotti: that's the tricky part. the ReportService.getStatus() is taking into account the end date compared with the today's date. Since the today's data is different every day, then the cached report status will get invalid. What I do to solve is storing without considering the dueDate, and re-add that in the query (indexed way) and also when re hydrating back

src/AppBundle/Service/ReportService.php
elvisciotti: and set as "_em " as a field property. I only added the missing declaration

src/AppBundle/Service/ReportService.php
elvisciotti: this code is not altering the logic, but only fixing the fact that the due date is not considered for the report status. I tested it carefully locally changing the due dates

src/AppBundle/Service/ReportService.php
elvisciotti: yes, i stopped refactor and didn't notice there were no dependencies anymore

elvisciotti: I moved it, pushed

src/AppBundle/Service/ReportService.php
elvisciotti: it's needed to be called after the array hydration. The reportStatusCache does not store the dueDate, but the client expects the report status to be readyToSubmit if also the date is due.

PR #481, opened 2018-11-23
[DO NOT MERGE] Ddpb 2338 documents bug

slizzio:


src/AppBundle/Service/ReportStatusService.php
```diff
@@ -432,8 +432,10 @@ public function getDocumentsState()
     {
         $numRecords = count($this->report->getDeputyDocuments());
 
-        if ($this->report->getWishToProvideDocumentation() === null || ($this->report->getWishToProvideDocumentation() === 'yes' && $numRecords == 0)) {
+        if ($this->report->getWishToProvideDocumentation() === null) {
             $status = ['state' => self::STATE_NOT_STARTED];
+        } elseif ($this->report->getWishToProvideDocumentation() === 'yes' && $numRecords == 0) {
```
elvisciotti: ok.alternative would be not storing the "yes" until the add the very first document. but the client atm stores the selection, so this is the easiest solution

slizzio: stores yes when adding a document but only deputy documents

PR #484, opened 2018-12-20
Ddpb 2343 deputy costs

elvisciotti:


src/AppBundle/Entity/Report/ProfDeputyInterimCost.php
```diff
@@ -0,0 +1,124 @@
...
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="IDENTITY")
+     * @ORM\SequenceGenerator(sequenceName="prof_deputy_interim_cost_id_seq", allocationSize=1, initialValue=1)
+     */
+    private $id;
+
+    /**
+     * @var Report
+     *
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Report\Report", inversedBy="profDeputyPreviousCosts")
```
slizzio: Does this inversedBy look correct?

src/AppBundle/Controller/Report/ProfDeputyPrevCostController.php
```diff
@@ -0,0 +1,136 @@
...
+        if (!array_key_exists('amount', $data)) {
+            throw new \InvalidArgumentException("Missing amount");
+        }
+        $cost = new EntityDir\Report\ProfDeputyPreviousCost($report, $data['amount']);
+        $this->updateEntity($data, $cost);
+        $report->setProfDeputyCostsHasPrevious('yes');
+        $this->persistAndFlush($cost);
+
+        $report->updateSectionsStatusCache($this->sectionIds);
+        $this->getEntityManager()->flush();
```
slizzio: Not sure the cache status should be updated before the data is flushed...?

src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -160,6 +160,42 @@ public function update(Request $request, $id)
...
+                        $profDeputyOtherCost->setMoreDetails($postedProfDeputyOtherCostType['more_details']);
+                    }
+
+                    $this->getEntityManager()->persist($profDeputyOtherCost);
+                }
+            }
+            $report->updateSectionsStatusCache([
+                Report::SECTION_PROF_DEPUTY_COSTS
+            ]);
+            $this->getEntityManager()->flush();
```
slizzio: Same here and a few other places : I would expect data to be flushed and then update cache but may not make a difference

src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -401,6 +437,94 @@ public function update(Request $request, $id)
...
+        }
+
+        if (array_key_exists('prof_deputy_costs_amount_to_scco', $data)) {
+            $report->setProfDeputyCostsAmountToScco($data['prof_deputy_costs_amount_to_scco']);
+            $report->updateSectionsStatusCache([Report::SECTION_PROF_DEPUTY_COSTS]);
+        }
+
+        if (array_key_exists('prof_deputy_costs_reason_beyond_estimate', $data)) {
+            $report->setProfDeputyCostsReasonBeyondEstimate($data['prof_deputy_costs_reason_beyond_estimate']);
+            $report->updateSectionsStatusCache([Report::SECTION_PROF_DEPUTY_COSTS]);
```
slizzio: Possibly move this to a single call towards the end of the method since cache status changing doesnt matter part way though the method

src/AppBundle/Controller/Report/ProfDeputyPrevCostController.php
aiesoftware: I don't think it matters because the `updateSectionsStatusCache` does not read the latest state from the database, but from the in memory entity object instead.

src/AppBundle/Controller/Report/ReportController.php
aiesoftware: I agree. It's probably just a safety mechanism to always update status cache as soon as a property is set on the entity which may alter the status of that section, just in case a new line is added afterwards which relies on the current status.

src/AppBundle/Controller/Report/ProfDeputyPrevCostController.php
slizzio: ok I wasn't sure, will leave as is

src/AppBundle/Controller/Report/ReportController.php
slizzio: will implement

src/AppBundle/Entity/Report/ProfDeputyInterimCost.php
slizzio: amended

PR #485, opened 2019-01-09
Ddpb 2357 amend report submission csv fields

aiesoftware:
**Changelog**
```
Created:
  AppBundle\Transformer\ReportSubmission\ReportSubmissionSummaryTransformer
    Responsible for transforming a collection of ReportSubmission’s into the suitable API response
  AppBundle\Service\DateTimeProvider
    Responsible for returning \DateTime objects in a manner that allows mocking of \DateTime objects in unit tests
    
Updated:
  AppBundle\Controller\Report\ReportSubmissionController
    Uses the new ReportSubmissionSummaryTransformer
  AppBundle\Entity\Repository\ReportSubmissionRepository
    Updates query to return Document data attached to Report

Removed:
  AppBundle\Service\ReportSubmissionStatService
    Replaced by ReportSubmissionSummaryTransformer
```
**PHPUnit**
```
AppBundle\Transformer\ReportSubmission\ReportSubmissionSummaryTransformer
 [x] Transforms a ReportSubmission
 [x] Ignores rows without a Report or NDR
 [x] Returns null document id if Report Document not found
```


src/AppBundle/Entity/Report/ReportSubmission.php
```diff
@@ -69,7 +69,7 @@ class ReportSubmission
      * @ORM\OneToMany(targetEntity="AppBundle\Entity\Report\Document", mappedBy="reportSubmission")
      * @ORM\JoinColumn(name="report_submission_id", referencedColumnName="id", onDelete="CASCADE")
      *
-     * @ORM\OrderBy({"createdBy"="ASC"})
+     * @ORM\OrderBy({"createdOn"="ASC"})
```
slizzio: Need to check this change doesnt affect Documents download page in admin. There may be a reason why they're ordered by CreatedBy and possibly cause a bug to be raised due to changed behaviour.

aiesoftware: Good spot - the change was not required. Have reverted

PR #491, opened 2019-02-01
DDPB 2427 user related page loads

aiesoftware:


src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -50,4 +50,50 @@ public function searchClients($query = '', $orderBy = 'lastname', $sortOrder = '
...
+     * @param User $user
+     * @param $clientId
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function saveUserToClient(User $user, $clientId)
+    {
+        $conn = $this->getEntityManager()->getConnection();
+
+        $conn->executeQuery(
+            'INSERT INTO deputy_case (client_id, user_id) VALUES (?, ?) ON CONFLICT DO NOTHING',
```
slizzio: would this insert duplicates?

aiesoftware: No, ON CONFLICT DO NOTHING handles it.

PR #495, opened 2019-02-11
Ddpb 2422 csv upload fix

slizzio:


src/AppBundle/Service/OrgService.php
```diff
@@ -257,20 +281,29 @@ private function upsertClientFromCsv(array $row, EntityDir\User $userOrgNamed)
...
+        foreach ($teams as $team) {
+            $members = $team->getMembers();
+            foreach ($members as $member) {
+                $client->addUser($member);
+                $depCount++;
             }
         }
+        $this->log('Assigned ' . $depCount . ' additional deputies to client');
+
+        $this->em->persist($client);
```
aiesoftware: This persist is only needed when the object has been created with `new` in this request - all objects retrieved from the repo are being managed by Doctrine by default - which is what `persist` does. Probably not a performance drain but good to know.

PR #505, opened 2019-03-04
Feature/DDPB-2457 - Add general management costs entity and migration

acsauk-mojd:


app/DoctrineMigrations/Version208.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
+        $this->addSql('DROP TABLE prof_deputy_management_cost');
+    }
+}
```
slizzio: this needs to be just a new field as part of report. Add the property to the estimates trait and it should just alter report table to add the new field

acsauk-mojd: Ahhh OK, will update!

src/AppBundle/Entity/Report/ProfDeputyManagementCost.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @var float
+     *
+     * @JMS\Type("float")
+     * @JMS\Groups({"prof-deputy-management-costs"})
+     *
+     * @ORM\Column(name="amount", type="float", precision=14, scale=2, nullable=true)
+     */
+    private $amount;
```
slizzio: this is the only property that needs adding to the trait (but renamed to give it meaning e.g. generalManagementCostAmount or something) re-use the same groups as the rest of the trait uses and it'll bring it back and persist

slizzio: Rest of the properties are not needed, ....sorry :)

acsauk-mojd: Gotcha - it did feel a bit strange duplicating all of that 🤦‍♂️

app/DoctrineMigrations/Version208.php
```diff
@@ -0,0 +1,32 @@
...
+class Version208 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE report ADD prof_dc_estimate_management_cost DOUBLE PRECISION DEFAULT NULL');
```
aiesoftware: For monetary amounts I believe we usually store as `NUMERIC(14, 2)` which is recommended for monetary fields.

src/AppBundle/Controller/Report/ReportController.php
```diff
@@ -537,6 +537,7 @@ public function update(Request $request, $id)
             $report->updateSectionsStatusCache([Report::SECTION_PROF_DEPUTY_COSTS]);
         }
 
+
```
aiesoftware: Unintended newline (-:

src/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandler.php
```diff
@@ -29,11 +29,30 @@ public function handle(Report $report, array $data)
...
+    /**
+     * @param Report $report
+     * @param array $data
+     * @throws \Doctrine\ORM\OptimisticLockException
+     */
+    private function updateManagementCost(Report $report, array $data) {
+        if (array_key_exists('prof_deputy_management_cost_amount', $data)) {
+            $report->setProfDeputyCostsEstimateManagementCostAmount($data['prof_deputy_management_cost_amount']);
+
+            $report->updateSectionsStatusCache([
```
aiesoftware: I don't _think_ this call is necessary as it is called at the end of all updates on line 36

src/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandler.php
```diff
@@ -29,11 +29,30 @@ public function handle(Report $report, array $data)
...
 
         $report->updateSectionsStatusCache([Report::SECTION_PROF_DEPUTY_COSTS_ESTIMATE]);
     }
 
+    /**
+     * @param Report $report
+     * @param array $data
+     * @throws \Doctrine\ORM\OptimisticLockException
+     */
+    private function updateManagementCost(Report $report, array $data) {
```
aiesoftware: I find it a good practice to have code methods declared in the order that they read - so I think this method should go after `updateMoreInfo` **or** if that calls any methods, then it goes after those methods.

src/AppBundle/Service/RestHandler/Report/ReportUpdateHandlerInterface.php
```diff
@@ -7,6 +7,8 @@
 interface ReportUpdateHandlerInterface
 {
     /**
+     * Handles form updates to persist changes throughout the flow.
```
aiesoftware: I try to avoid comments at all cost because we can never trust them. Human language is ambiguous whereas code is not, but mainly we have no idea if the comment is still true, because when code changes, comments never do. If a comment is needed it's usually a sign that methods, classes or variables could have better names. In this case, the interface doesn't care that the data comes from a form, just an array - the api doesn't deal with forms per se, just data. Personally I'd remove it and hope that the code speaks for itself here (-:

tests/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandlerTest.php
```diff
@@ -225,4 +233,12 @@ private function assertNewProfDeputyEstimateCostIsCreated()
         $this->assertEquals(false, $profDeputyEstimateCost->getHasMoreDetails());
         $this->assertEquals(null, $profDeputyEstimateCost->getMoreDetails());
     }
+
+    public function costEstimateDataProvider()
```
aiesoftware: Handy little dataProvider. Can it live underneath the test to which it applies please?

src/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandler.php
acsauk-mojd: Ah yeah good spot - I'll drop 👍

src/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandler.php
acsauk-mojd: That makes sense - I think I'd unconsciously been doing this for a while but not realising. I'll move this down below `updateMoreInfo`.

src/AppBundle/Service/RestHandler/Report/ReportUpdateHandlerInterface.php
acsauk-mojd: Agree! I'll remove.

src/AppBundle/Entity/Report/Report.php
```diff
@@ -441,8 +441,8 @@ public function __construct(Client $client, $type, \DateTime $startDate, \DateTi
         }
         $this->type = $type;
         $this->client = $client;
-        $this->startDate = new \DateTime($startDate->format('Y-m-d'));
-        $this->endDate = new \DateTime($endDate->format('Y-m-d'));
+        $this->startDate = new \DateTime($startDate->format('Y-m-d'), new \DateTimeZone('Europe/London'));
+        $this->endDate = new \DateTime($endDate->format('Y-m-d'), new \DateTimeZone('Europe/London'));
```
slizzio: ?

tests/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandlerTest.php
```diff
@@ -26,8 +26,9 @@ class DeputyCostsEstimateReportUpdateHandlerTest extends TestCase
      */
     public function setUp()
     {
+        $date = new \DateTime('now', new \DateTimeZone('Europe/London'));
         $this->report = $this->getMockBuilder(Report::class)
-            ->setConstructorArgs([new Client, Report::TYPE_102, new \DateTime, new \DateTime])
+            ->setConstructorArgs([new Client, Report::TYPE_102, $date, $date])
```
slizzio: I think you need to refresh your branch with latest commits?

src/AppBundle/Entity/Report/Report.php
acsauk-mojd: I was getting test failure where it complained about using default timezones so I set them here and in the test.

tests/AppBundle/Service/RestHandler/Report/DeputyCostsEstimateReportUpdateHandlerTest.php
acsauk-mojd: I've refreshed with `feature-105` - this was related with failures during test runs due to not explicitly setting a timezone.

PR #506, opened 2019-03-05
Ddpb 2455 show ndr reports 2

slizzio:


src/AppBundle/Entity/Client.php
```diff
@@ -828,7 +828,7 @@ public function setClientContacts($clientContacts)
      */
     public static function isValidCaseNumber($query)
     {
-        return (bool) preg_match('/^[0-9t]{8}$/i', $query);
+        return (bool) (preg_match('/^[0-9t]{8}$/i', $query) || substr($query, 0, 5) == 'behat');
```
aiesoftware: Is this needed?

slizzio: Removed and amended tests

PR #507, opened 2019-03-06
DDPB-2463 Page timeout on user edit page

aiesoftware:
**DO NOT MERGE**
**Addresses:**
* Poor performance for pages with high data volumes due to Doctrine hydration and JMS Serialisation
* Gigantic payloads being sent to frontend Client
* Complex nature of JMS serialisation groups in attempt to reduce payloads, but they are confusing, lack consistency and are an attempt to fix the issue of how can we serialise our entities and all of their complex relationships.

**Proposed solutions**
* Create clear boundaries between what the API knows and what the Client _needs_ to know, using Data Transfer Objects (DTO's) to represent trimmed down versions of our Entities to communicate across the wire
* Avoid unnecessary Doctrine hydration and entity serialisation by using Repositories to write native SQL, and using Assemblers to create DTO's from SQL results. Use Controllers to ask a Repository for a DTO and return a json encoded DTO in the response

**Notes**
* There is still a performance hit on the Client side as it currently still hydrates the response into full blown Entities. We should work towards using DTO's on both sides to further secure the boundary between the Application layer and the Domain layer
* New code in the API has been put in a `v2` folder and namespace, and new endpoint is prefixed with `/v2`

**Result**
* On local dev environment, it took `admin/edit-user` for a user with 950 clients to display from ~37 secs. to ~7 secs. Reduces down to 3 secs. when using DTO on Client side (experimental)

**PHPUnit output**
```
AppBundle\v2\Assembler\ClientAssembler
 [x] assembleFromArray returns DTO from given array
 [x] assembleFromArray throws exception if given insufficient data

AppBundle\v2\Assembler\DeputyAssembler
 [x] assembleFromArray returns DTO from given array
 [x] assembleFromArraythrows exception if given insufficient data

AppBundle\v2\Repository\ClientRepository
 [x] getDtoDataArrayByDeputy returns client data for each client belonging to deputy

AppBundle\v2\Repository\DeputyRepository
 [x] getDtoById returns deputy DTO for given deputy
```


app/config/services/repositories.yml
```diff
@@ -0,0 +1,26 @@
...
+  UserRepository:
+    class: AppBundle\Entity\User
+    factory: ["@doctrine.orm.entity_manager", getRepository]
+    arguments:
+      - AppBundle\Entity\User
+
+  AppBundle\v2\Repository\DeputyRepository:
+    class: AppBundle\v2\Repository\DeputyRepository
+    arguments:
+      - '@doctrine.orm.entity_manager'
```
slizzio: If the Repository extends AbstractEntityRepository then it has the entity manager already injected via $this->_em

app/config/services/repositories.yml
```diff
@@ -0,0 +1,26 @@
...
+      - '@UserRepository'
+      - '@AppBundle\v2\Repository\ClientRepository'
```
slizzio: comment only: Not convinced we need to inject this. AbstractEntityRepository supplies methods to get other repositories but appreciate the need to inject to allow easier unit testing so happy to leave it in. Just conscious that as the repositories grow in complexity you may find developers adding several other repositories to the constructor as they see fit for their needs

src/AppBundle/Entity/User.php
```diff
@@ -12,7 +12,7 @@
  * Users.
  *
  * @ORM\Table(name="dd_user", indexes={@ORM\Index(name="deputy_no_idx", columns={"deputy_no"})})
- * @ORM\Entity(repositoryClass="AppBundle\Entity\Repository\UserRepository")
```
slizzio: why is this removed? (curiosity)

app/config/services/repositories.yml
aiesoftware: Technically yes, but composition should be favoured over inheritance. Extending a class (which is already an extension) just to get at a property within a parent is a code smell and probably means we should inject the service we need.

src/AppBundle/v2/Assembler/ClientAssembler.php
```diff
@@ -0,0 +1,53 @@
...
+
+class ClientAssembler
+{
+    /**
+     * @param array $data
+     * @return ClientDto
+     */
+    public function assembleFromArray(array $data)
+    {
+        $this->throwExceptionIfMissingRequiredData($data);
```
slizzio: Maybe 
if (!$this->dataIsValid($data)) {           
 throw new \InvalidArgumentException(__CLASS__ . ': Missing all data required to build DTO');
}
or even rename dataIsValid() to validateData() and put it in an interface? to ensure all future assemblers have that method?

app/config/services/repositories.yml
aiesoftware: Unit testing is a very valid reason in my opinion. If this repository happens to need another repository, then I think it would be a valid reason to inject one.

src/AppBundle/v2/Assembler/ClientAssembler.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return
+            array_key_exists('id', $data) &&
+            array_key_exists('case_number', $data) &&
+            array_key_exists('firstname', $data) &&
+            array_key_exists('lastname', $data) &&
+            array_key_exists('email', $data) &&
+            array_key_exists('report_count', $data) &&
+            array_key_exists('ndr_id', $data);
+    }
```
slizzio: (optional) consider using 
   return !empty(array_diff(['id', 'case_number' ... ,'ndr_id'], $data));

src/AppBundle/Entity/User.php
aiesoftware: One of the benefits of making the Repository classes standalone, without extending Doctrine, is that we can have many Repo's for one Entity, and therefore don't need to declare this here.

src/AppBundle/v2/Assembler/ClientAssembler.php
slizzio: (optional) consider moving this to an Abstract class that references self::dtoKeys set as a property ?

src/AppBundle/v2/Assembler/ClientAssembler.php
aiesoftware: Using the software rule of 3, I'd rather not break anything out into abstract classes or interface methods yet. Your top suggestion is pretty much what is happening inside that method.

src/AppBundle/v2/Assembler/DeputyAssembler.php
```diff
@@ -0,0 +1,107 @@
...
+            array_key_exists('address_postcode', $data) &&
+            array_key_exists('odr_enabled', $data) &&
+            array_key_exists('clients', $data);
+    }
+
+    /**
+     * @param $deputy
+     * @return DeputyAssembler
+     */
+    private function buildDeputyDtoFromArray($deputy)
```
slizzio: if we lose the 'Deputy' from this method it will allow much easier refactoring later on

aiesoftware: Okay

src/AppBundle/v2/Assembler/DeputyAssembler.php
```diff
@@ -0,0 +1,107 @@
...
+        );
+
+        return $this;
+    }
+
+    /**
+     * @param array $clients
+     * @return DeputyAssembler
+     */
+    private function buildAndAttachClientDtosFromArray(array $clients)
```
slizzio: same here. Should just be 'buildDTO()'

src/AppBundle/v2/Assembler/ClientAssembler.php
aiesoftware: Would this care about the order of keys? The method doesn't care about order, just existence..

src/AppBundle/v2/Assembler/DeputyAssembler.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $this->deputyDto->setClients($clients);
+
+        return $this;
+    }
+
+    /**
+     * @return DeputyDto
+     */
+    private function getDeputyDto()
```
slizzio: getDto()

src/AppBundle/v2/Assembler/DeputyAssembler.php
aiesoftware: This method name tells a story, it helps devs understand what is happening without having to read inside the method.

src/AppBundle/v2/Assembler/DeputyAssembler.php
aiesoftware: Okay.

src/AppBundle/v2/Controller/DeputyController.php
```diff
@@ -0,0 +1,53 @@
...
+     * @Route("/{id}", requirements={"id":"\d+"})
+     * @Method({"GET"})
+     *
+     * @param $id
+     * @return JsonResponse
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function getByIdAction($id)
+    {
+        return $this->buildSuccessResponse($this->repository->getDtoById($id));
```
slizzio: This assumes success

src/AppBundle/v2/Assembler/ClientAssembler.php
aiesoftware: Not keen on forcing the abstraction yet.

src/AppBundle/v2/Controller/DeputyController.php
aiesoftware: I'll rephrase or reorder it 👍

src/AppBundle/v2/Controller/DeputyController.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @param DeputyDto $dto
+     * @return JsonResponse
+     */
+    private function buildSuccessResponse(DeputyDto $dto)
+    {
+        return new JsonResponse([
+            'success' => true,
+            'data' => $dto->jsonSerialize(),
```
slizzio: Could the Dto not serialize the data before returning?

src/AppBundle/v2/DTO/ClientDto.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * @param $id
+     * @param $caseNumber
+     * @param $firstName
+     * @param $lastName
+     * @param $email
+     * @param $reportCount
+     * @param $ndrId
+     */
+    public function __construct($id, $caseNumber, $firstName, $lastName, $email, $reportCount, $ndrId)
```
slizzio: would it be better to have the last two parameters optional?

src/AppBundle/v2/Controller/DeputyController.php
aiesoftware: A DTO is an object which will need to be passed through the application as a DTO, not a serialized representation. Make sense?

src/AppBundle/v2/DTO/ClientDto.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->id,
+            'case_number' => $this->caseNumber,
+            'firstname' => $this->firstName,
+            'lastname' => $this->lastName,
+            'email' => $this->email,
+            'total_report_count' => $this->reportCount,
```
slizzio: id leave as just report_count

aiesoftware: Needs to match the entity on the client side unfortunately

src/AppBundle/v2/DTO/ClientDto.php
aiesoftware: Maybe, but not sure. Leave it for now?

src/AppBundle/v2/DTO/DeputyDto.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    public function jsonSerialize()
+    {
+        return [
+            'id' => $this->id,
+            'firstname' => $this->firstName,
+            'lastname' => $this->lastName,
+            'email' => $this->email,
+            'role_name' => $this->roleName,
+            'address_postcode' => $this->postcode,
```
slizzio: again, if we keep consistent naming conventions will make it easier to refactor:
address_postcode => $this->addressPostcode

aiesoftware: See above

src/AppBundle/v2/DTO/DeputyDto.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * @return array
+     */
+    private function serializeClients()
+    {
+        $serializedClients = [];
+
+        foreach ($this->clients as $client) {
```
slizzio: what happens if clients is null?

src/AppBundle/v2/DTO/DeputyDto.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        return $this->firstName;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getLastName()
+    {
+        return $this->lastName;
```
slizzio: check database fields I think its lastname and firstname, no capital N

src/AppBundle/v2/DTO/DeputyDto.php
aiesoftware: Good spot, I'll verify it

src/AppBundle/v2/Repository/ClientRepository.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->entityManager = $entityManager;
+        $this->clientRepository = $clientRepository;
+    }
+
+    /**
+     * @param $deputyId
+     * @return array
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public function getDtoDataArrayByDeputy($deputyId)
```
slizzio: getDtoData()  then we can put in an interface

src/AppBundle/v2/DTO/DeputyDto.php
aiesoftware: Do we care about it matching database?

src/AppBundle/v2/Repository/DeputyRepository.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace AppBundle\v2\Repository;
+
+use AppBundle\v2\Assembler\DeputyAssembler;
+use Doctrine\Common\Persistence\ObjectRepository;
+use Doctrine\ORM\EntityManagerInterface;
+
+class DeputyRepository
```
slizzio: extends AbstractRepository

src/AppBundle/v2/Repository/DeputyRepository.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    /** @var EntityManagerInterface */
+    private $entityManager;
```
slizzio: not needed use $this->_em

src/AppBundle/v2/Repository/DeputyRepository.php
aiesoftware: See previous comments on that

PR #510, opened 2019-03-08
DDPB-2390 Fix unit tests with relative years

gregtyler:
The `getExpectedReportStartDate` and `getExpectedReportEndDate` methods use `date('Y')` to calculate their output, meaning the tests are all relative to the current year. We want static test input and output.

Because `date()` is tightly integrated into the functions, we can't easily create static tests. To get around this, this PR adds an additional `$year` argument to the methods which is used in the calculations instead. This can be manually specified, but defaults to `date('Y')` so existing functionality isn't impeded.

The result allows for static unit tests, but adds the overhead of an additional function argument which is only used in testing.

### DIFF
--- a/src/AppBundle/Entity/Client.php
+++ b/src/AppBundle/Entity/Client.php
@@ -897,8 +897,13 @@ class Client
      * @JMS\Groups({"checklist-information"})
      * @return \DateTime|null
      */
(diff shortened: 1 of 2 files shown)


gregtyler:
This PR is no longer needed now that #513 has merged the commits into `master`

PR #514, opened 2019-03-11
Ddpb 2463 api timeouts on various pages

aiesoftware:


src/AppBundle/v2/Assembler/ClientAssembler.php
```diff
@@ -2,52 +2,92 @@
...
-        return new ClientDto(
-            $data['id'],
-            $data['case_number'],
-            $data['firstname'],
-            $data['lastname'],
-            $data['email'],
-            $data['report_count'],
-            $data['ndr_id']
-        );
+        $dto = new ClientDto();
```
slizzio: Just out of curiousity, why remove these params from the constructor and have the object constructed outside the class?

src/AppBundle/v2/Assembler/Report/ReportAssemblerFromEntityDecorator.php
```diff
@@ -0,0 +1,66 @@
...
+    private $decorated;
+
+    /** @var StatusAssembler  */
+    private $statusDtoAssembler;
+
+    /** @var ReportRepository */
+    private $reportRepository;
+
+    /**
+     * @param ReportAssemblerInterface $decorated
```
slizzio: I think an explanation on what decorated means in this context

src/AppBundle/v2/Transformer/ClientTransformer.php
```diff
@@ -0,0 +1,94 @@
...
+            'email' => $dto->getEmail(),
+            'archived_at' => $this->transformArchivedAt($dto),
+            'total_report_count' => $dto->getReportCount()
+        ];
+
+        if (!in_array('reports', $exclude)) {
+            $transformed['reports'] = $this->transformReports($dto->getReports());
+        }
+
+        if (!in_array('ndr', $exclude) && $dto->getNdr() instanceof NdrDto) {
```
slizzio: is this right? $dto->getNdr() instanceof NdrDto
It could be, just checking we are dealing with Dto's  and not the NDR itsself

PR #525, opened 2019-03-20
 DDPB-2517: Allow more complex users in test fixtures

gregtyler:
In DDPB-2517, we are adding new user accounts to test fixtures to allow testers to easily log in as professional and PA deputies.

Supporting these new deputy types requires some changes to the `AddSingleUserCommand` command:

* Assigning deputies to teams
* Allowing multiple deputies to be assigned to the same client
* Automatically generating reports for professional and PA deputies (because that drives the dashboard)

### DIFF
--- a/src/AppBundle/Command/AddSingleUserCommand.php
+++ b/src/AppBundle/Command/AddSingleUserCommand.php
@@ -160,8 +160,17 @@ class AddSingleUserCommand extends ContainerAwareCommand
         /**
          * Create report
          */
-        if (in_array($user->getRoleName(), [User::ROLE_PROF_NAMED, User::ROLE_PA_NAMED])) {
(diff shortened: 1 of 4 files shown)


src/AppBundle/Command/AddSingleUserCommand.php
```diff
@@ -137,8 +141,49 @@ protected function addSingleUser(OutputInterface $output, array $data, array $op
                 $ndr = new Ndr($client);
                 $em->persist($ndr);
             }
+        } else if (isset($data['caseNumber'])) {
+            // If client already exists, just assign user
+            $client = $clientRepo->findOneBy(['caseNumber' => CasRec::normaliseCaseNumber($data['caseNumber'])]);
+            $user->addClient($client);
```
aiesoftware: Worth clarifying that `$client instanceof Client`

PR #526, opened 2019-03-25
DDPB-2512 - NDR report incorrectly displayed

acsauk-mojd:
I've moved the NDR creation step to the Lay Deputy specific steps during the upsert action to ensure they are only created for Lay Deputies.

As I was running the unit tests locally I encountered an error while loading the fixtures class so I've namespaced this and updated references as part of this PR.

### DIFF
--- a/src/AppBundle/Controller/ClientController.php
+++ b/src/AppBundle/Controller/ClientController.php
@@ -67,7 +67,8 @@ class ClientController extends RestController
             $client->setDateOfBirth($dob);
         }
 
-        $this->persistAndFlush($client);
+        $this->getEntityManager()->persist($client);
+        $this->getEntityManager()->flush();
 
         return ['id' => $client->getId()];
     }
--- a/src/AppBundle/Controller/ClientController.php
+++ b/src/AppBundle/Controller/ClientController.php
@@ -51,7 +51,7 @@ class ClientController extends RestController
         ]);
 
         if ($user && $user->isLayDeputy()) {
-            if (!$client->getNdr() && $user->getNdrEnabled()) {
+            if (array_key_exists('ndr_enabled', $data) && $data['ndr_enabled'] && !$client->getNdr()) {
                 $ndr = new EntityDir\Ndr\Ndr($client);
                 $em->persist($ndr);
             }
--- a/src/AppBundle/Controller/ClientController.php
+++ b/src/AppBundle/Controller/ClientController.php
@@ -49,6 +49,13 @@ class ClientController extends RestController
         ]);
 
         if ($this->getUser()->isLayDeputy()) {
+            //add NDR if not added yet
+            // TODO move to listener or service
+            if (!$client->getNdr()) {
+                $ndr = new EntityDir\Ndr\Ndr($client);
+                $this->getEntityManager()->persist($ndr);
+            }
+
             $client->setCourtDate(new \DateTime($data['court_date']));
             $this->hydrateEntityWithArrayData($client, $data, [
                 'case_number' => 'setCaseNumber',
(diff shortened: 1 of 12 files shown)


src/AppBundle/Controller/ClientController.php
```diff
@@ -49,6 +49,13 @@ public function upsertAction(Request $request)
         ]);
 
         if ($this->getUser()->isLayDeputy()) {
+            //add NDR if not added yet
+            // TODO move to listener or service
```
aiesoftware: I think we should remove these comments..

acsauk-mojd:
The requirements of this ticket evolved slightly - NDRs should only be created for Lays if the NDR Enabled checkbox has been ticked from the admin side. We were previously just checking if an NDR existed for the client rather than checking for the box being checked. I've updated the logic to reflect this and added a couple fo extra assertions for this in the unit tests.

src/AppBundle/Controller/ClientController.php
```diff
@@ -60,14 +71,8 @@ public function upsertAction(Request $request)
...
 
-        $this->persistAndFlush($client);
-
-        //add NDR if not added yet
-        // TODO move to listener or service
-        if (!$client->getNdr()) {
-            $ndr = new EntityDir\Ndr\Ndr($client);
-            $this->persistAndFlush($ndr);
-        }
+        $em->persist($client);
```
slizzio: check client is set before persisting

tests/AppBundle/Controller/ClientControllerTest.php
```diff
@@ -121,10 +121,9 @@ public function testupsertPostLayDeputy()
         $this->assertEquals('Firstname', $client->getFirstname());
         $this->assertCount(1, $client->getUsers());
         $this->assertEquals(self::$deputy1->getId(), $client->getUsers()->first()->getId());
-        $this->assertInstanceOf('AppBundle\Entity\Ndr\Ndr', $client->getNdr());
     }
 
-    public function testupsertPut()
+    public function testupsertPut_lay_deputy()
```
slizzio: camelCase?

tests/AppBundle/Controller/ClientControllerTest.php
```diff
@@ -148,6 +147,30 @@ public function testupsertPut()
         $this->assertEquals(null, $client->getDateOfBirth());
         $this->assertEquals('2015-12-31', $client->getCourtDate()->format('Y-m-d'));
         $this->assertEquals(self::$deputy1->getId(), $client->getUsers()->first()->getId());
+    }
+
+    public function testupsertPut_lay_deputy_ndr_enabled()
```
slizzio: camelCase

tests/AppBundle/Controller/ClientControllerTest.php
```diff
@@ -148,6 +147,30 @@ public function testupsertPut()
...
+            'mustSucceed' => true,
+            'AuthToken' => self::$tokenDeputy,
+            'data' => ['id' => self::$client1->getId(), 'ndr_enabled' => true] + $this->updateDataLay,
+        ]);
+        self::fixtures()->clear();
+        $client = self::fixtures()->getRepo('Client')->find($return['data']['id']); /* @var $client \AppBundle\Entity\Client */
+        $this->assertInstanceOf(Ndr::class, $client->getNdr());
+    }
+
+    public function testupsertPut_PA()
```
slizzio: camelcase

tests/AppBundle/Controller/ClientControllerTest.php
acsauk-mojd: I used to be exclusively camel case for test function names until my last role where we kept the actual function it was testing in camel case but gave the scenario/detail using snake case. Happy to change if thats the preferred style here though!

PR #532, opened 2019-04-05
DDPB-2539 report submission registration stats

acsauk-mojd:
Controller could probably be refactored to use a service class and some testing is needed but I've run out of time 😭 sorry guys!


src/AppBundle/Service/StatsService.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    /** @var EntityManager */
+    private $em;
+
+    public function __construct($em)
+    {
+        $this->em = $em;
+    }
+
+    public function countNamedDeputies($type, $from, $to)
```
slizzio: we use type quite extensively in the api, perhaps 'role' would make more sense?

slizzio: and maybe force from and to to be \DateTime 's in case the input changes to a date that doesnt resolve

src/AppBundle/Service/StatsService.php
```diff
@@ -0,0 +1,54 @@
...
+                ->andWhere('u.registrationDate BETWEEN :from AND :to')
+                ->setParameters(['from' => $from, 'to' => $to, 'type' => $type])
+                ->getQuery()
+                ->getSingleScalarResult();
+        } catch(NoResultException $e) {
+            return 0;
+        }
+    }
+
+    public function countReports($from, $to)
```
slizzio: and here too?     public function countReports(\Datetime $from, \Datetime $to)

src/AppBundle/Service/StatsService.php
gregtyler: Good points, both done 👍

gregtyler:
Closed pending story refinement

PR #533, opened 2019-04-05
DDPB-2524: Add more fixture users

gregtyler:
This moves fixture user creation to the Doctrine Fixtures package, and removes the reliance on environment variables to define those users.


src/AppBundle/DataFixtures/TeamFixtures.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $manager->flush();
+    }
+
+    private function createTeam($code, $name, $manager) {
+        // Create team
+        $team = new Team($name . ' Team');
+        $manager->persist($team);
+
+        // Create team admin user
```
aiesoftware: Whenever a one line comment can nicely describe the next few lines, it's usually a candidate to create a method with the same name as that one line. I'm a big fan of writing descriptive code rather than untrustworthy comments. Not a blocker as this is just fixture code, just a comment.

src/AppBundle/DataFixtures/TeamFixtures.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function getOrder()
+    {
+        return 10;
+    }
+
+    protected function getEnvironments()
+    {
+        return array('dev');
```
aiesoftware: Not a blocker: `return ['dev']`

src/AppBundle/DataFixtures/TeamFixtures.php
gregtyler: Yeah, that's definitely good advice. I'll bear it in mind for the future, but I don't think it'll make a great difference here because the duplication is quite minimal.

PR #536, opened 2019-04-15
DDBP-2524: Load fixture parameter on demand

gregtyler:
Removing auto-wiring means that you don't have to define the fixture parameter to start the application. Environments which define the parameter can load fixtures, but those which don't can't.


gregtyler:
Self-merging, since this is a config change for a live issue

PR #546, opened 2019-05-20
DDPB-2605: Removes 'other' option from breakdown and removes getProfD…

aiesoftware:
…eputyEstimateCostsTotal() method.

Neither are no longer required as a result of https://opgtransform.atlassian.net/browse/DDPB-2605 and https://opgtransform.atlassian.net/browse/DDPB-2606


src/AppBundle/Entity/Report/Traits/ReportProfDeputyCostsEstimateTrait.php
```diff
@@ -63,8 +63,7 @@ trait ReportProfDeputyCostsEstimateTrait
         ['typeId' => 'contact-client', 'hasMoreDetails' => false],
         ['typeId' => 'contact-case-manager-carers', 'hasMoreDetails' => false],
         ['typeId' => 'contact-others', 'hasMoreDetails' => false],
-        ['typeId' => 'forms-documents', 'hasMoreDetails' => false],
-        ['typeId' => 'other', 'hasMoreDetails' => true],
```
slizzio: Do we need any migrations to remove these costs from the database?

aiesoftware: Yes, good spot. We shall need to remove from db on release.

PR #551, opened 2019-05-28
Don't add `CREATE SCHEMA public` commands to down migrations

gregtyler:
## Purpose
Down migrations generated by Doctrine include a `CREATE SCHEMA public` command. This is erroneous and causes the down migrations to fail (because the public schema already exists).

This is [a known issue](https://github.com/doctrine/dbal/issues/1110) in Doctrine DBAL.

## Approach
I removed the `CREATE SCHEMA public` command from existing migrations, and also introduced a listener which stops them from being added unless necessary (e.g. if we start using multiple schemas).

## Learning
I took the fix from [a comment on the GitHub issue](https://github.com/doctrine/dbal/issues/1110#issuecomment-255765189). This seems to be the accepted solution for now, awaiting Doctrine to fix the issue permanently.

## Checklist
- [N/A] I have updated documentation (Confluence/GitHub wiki) where relevant
- [N/A] I have added tests to prove my work, and they follow our [best practices](https://github.com/ministryofjustice/opg-digi-deps-client/wiki/Testing-best-practices)
- [N/A] I have successfully built my branch to a feature environment
  - False positive on Symfony error fixed in #549
- [N/A] The product team have tested these changes

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -1,33 +1,15 @@
 imports:
     - { resource: services/assemblers.yml }
     - { resource: services/controllers.yml }
+    - { resource: services/event_listeners.yml }
     - { resource: services/rest_handlers.yml }
     - { resource: services/repositories.yml }
     - { resource: services/transformers.yml }
 
 services:
-    kernel.listener.responseConverter:
-        class: AppBundle\EventListener\RestInputOuputFormatter
-        arguments: [ "@jms_serializer", "@logger", ["json"], "json", "%kernel.debug%" ]
-        public: true
-        tags:
-            - { name: kernel.event_listener, event: kernel.view, method: onKernelView }
-            - { name: kernel.event_listener, event: kernel.exception, method: onKernelException }
-            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
     em:
         alias: doctrine.orm.entity_manager
         public: true
-
-    doctrine.listener:
-        class: AppBundle\EventListener\DoctrineListener
-        tags:
-            - { name: doctrine.event_listener, event: prePersist, method: prePersist }
-            - { name: doctrine.event_listener, event: preRemove, method: preRemove }
-
-    fixDefaultSchemaListener:
-        class: AppBundle\EventListener\FixDefaultSchemaListener
-        tags:
-            - { name: doctrine.event_listener, event: postGenerateSchema, method: postGenerateSchema }
 
     opg_digideps.casrec_verification_service:
         class: AppBundle\Service\CasrecVerificationService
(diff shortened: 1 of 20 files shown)


src/AppBundle/EventListener/FixDefaultSchemaListener.php
```diff
@@ -0,0 +1,25 @@
...
+namespace AppBundle\EventListener;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;
+
+class FixDefaultSchemaListener implements EventSubscriber
+{
+    public function getSubscribedEvents()
+    {
+        return array(
```
aiesoftware: For consistency can we use `[]` instead of `array` please.

src/AppBundle/EventListener/FixDefaultSchemaListener.php
```diff
@@ -0,0 +1,25 @@
...
+            'postGenerateSchema',
+        );
+    }
+
+    public function postGenerateSchema(GenerateSchemaEventArgs $Args)
+    {
+        $Schema = $Args->getSchema();
```
aiesoftware: Can we lowercase these two vars please?

app/config/services.yml
```diff
@@ -24,6 +24,11 @@ services:
             - { name: doctrine.event_listener, event: prePersist, method: prePersist }
             - { name: doctrine.event_listener, event: preRemove, method: preRemove }
 
+    fixDefaultSchemaListener:
+        class: AppBundle\EventListener\FixDefaultSchemaListener
+        tags:
+            - { name: doctrine.event_listener, event: postGenerateSchema, method: postGenerateSchema }
+
```
aiesoftware: I think it's good to avoid a giant `services.yml`. You'll see I've gradually created service files in the `app/services` directory, and import each one from `services.yml`. It might be a good idea to create an `event_listeners.yml`.

Also, in keeping with the naming convention of the services in those new files, we could name this service `AppBundle\EventListener\FixDefaultSchemaListener` which I think also helps with auto wiring (although not strictly relevant here).

src/AppBundle/EventListener/FixDefaultSchemaListener.php
gregtyler: Done 👍

src/AppBundle/EventListener/FixDefaultSchemaListener.php
gregtyler: Done 👍

app/config/services.yml
gregtyler: Good idea; I moved the other listeners in there too 👍

PR #560, opened 2019-06-24
DDPB-2629: Upgrade PHP to 7.3

gregtyler:
## Purpose
PHP 5.x has reached EOL, so we need to upgrade to PHP 7.

Fixes [DDPB-2629](https://opgtransform.atlassian.net/browse/DDPB-2629)

## Approach
The fundamental changes here are pretty straightforward: use PHP7 base images in the Dockerfile and simplify the extension configuration now that we can use the latest versions.

A few other changes:
- Upgraded `doctrine/orm` since ≤2.5 is only compatible with PHP5 and ≥2.6 is only compatible with PHP7
- Removed `zend_extension` configuration strings since these are done as part of `docker-php-ext-install` and `docker-php-ext-enable`
- Removed an unsupported argument to `time()` (PHP is fussy about this now)
- Replaced `catch (\Exception)` with `catch (\Throwable)` to [respect changes in PHP7](https://www.php.net/manual/en/language.errors.php7.php)

## Learning
We should review our other dependencies, since it's now possible to upgrade a lot of them to more recent versions which will be faster and more secure.

## Checklist
- [ ] I have updated documentation (Confluence/GitHub wiki) where relevant

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     }
   },
   "require": {
+    "php": ">=7.3.5",
     "symfony/symfony": "3.4.*",
     "doctrine/orm": "2.*",
     "doctrine/doctrine-bundle": "1.9.*",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
   },
   "require": {
     "symfony/symfony": "3.4.*",
-    "doctrine/orm": "2.5.*",
+    "doctrine/orm": "2.*",
     "doctrine/doctrine-bundle": "1.9.*",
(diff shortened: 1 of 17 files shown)


composer.json
```diff
@@ -17,9 +17,8 @@
     }
   },
   "require": {
-    "php": ">=5.5.9",
```
aiesoftware: What's the reason we are removing instead of replacing the version?

gregtyler: Good point, I got confused what I was removing there

PR #566, opened 2019-07-08
Ddpb 2699 lay csv refactor

aiesoftware:
## Purpose
```
As a system
In order to prevent corrupt updates and insertions
I should not process CSV rows whose case number belongs to a different client/deputy pairing 
```

Fixes [DDPB-2699](https://opgtransform.atlassian.net/browse/DDPB-2699)

## Approach

Rewrote new code to handle the CSV upload under the `v2` directory, in a `Registration` domain.
The main goals of the refactor:
* Fix the bug that is incorrectly updating certain report types
* Abstract the code into services that can be unit tested in isolation
* Improve unit test coverage of the upload process
* Create a fixed boundary into the `Registration` domain by introducing a common `LayDeputyshipDto` and an interface that is responsible for taking an input from any source and converting it into a collection of `LayDeputyshipDto` objects. This better paves the way for supporting uploads from other sources (Sirius) by extending code rather than modifying existing code.

For an upload of Lay Deputyship data into the API, we have the following components:
* A new `Controller` to receive compressed data 
* A new `DataCompression` service to decompress the data (extracted from `CsvUploader` and deprecated)
* A CasRec specific `LayDeputyshipDtoAssemblerInterface` implementation to convert the raw CasRec data into a collection of `LayDeputyshipDto` objects
* A `LayDeputyshipUploader` to take the `LayDeputyshipDto` objects and process according to our registration needs.

To handle an upload from another source, we create a new implementation of `LayDeputyshipDtoAssemblerInterface` that will convert the raw data into a collection of `LayDeputyshipDto` objects. We can then send that into the `LayDeputyshipUploader` to process in the same way.

## Learning
Until we upgrade our version of PHPUnit, we are unable to declare return types for methods that are being mocked in unit tests.

## PHPUnit output
```
s\AppBundle\v2\Registration\Uploader\LayDeputyshipUploader
 [x] Throws exception if data set too large
 [x] Ignores deputyships where client already registered to another deputy
 [x] Persists an entry for each valid deputyship
 [x] Ignores deputyships with invalid deputyship data

s\AppBundle\v2\Registration\SelfRegistration\CasRecFactory
 [x] Throws exception if creates invalid entity
 [x] Returns a valid hydrated CasRec entity

s\AppBundle\v2\Registration\Assembler\LayDeputyshipDtoCollectionAssembler
 [x] assembleFromArray throws exception if item contains missing data
 [x] assembleFromArray assembles a collection and returns it

s\AppBundle\v2\Registration\Assembler\LayDeputyshipDtoAssembler
 [x] canAssemble returns true if given complete data
 [x] canAssemble returns false if given incomplete data
 [x] assembleFromArray assembles and returns a lay deputyship dto
 [x] assembleFromArray determines if ndr enabled
```

## Checklist


src/AppBundle/Entity/Repository/ClientRepository.php
```diff
@@ -112,4 +112,28 @@ public function getArrayById($id)
...
+                    inner join deputy_case dc on c.id = dc.client_id 
+                    inner join dd_user u on dc.user_id = u.id 
+                    where c.case_number = ? 
+                    and u.deputy_no != ? 
+                    and u.deputy_no is not null',
+            [$caseNumber, $deputyNumber]
+        );
+
+        // Result is either false for no match, or an array result for the different deputy that is assigned.
+        return $stmt->fetch();
```
gregtyler: Based on the name and annotation, this function should explicitly return a boolean rather than an array. Just a `!!` on return would probably do it?

src/AppBundle/Service/DataNormaliser.php
```diff
@@ -0,0 +1,74 @@
...
+    public function normaliseSurname(string $value): string
+    {
+        $value = trim($value);
+        $value = strtolower($value);
+        $value = strtr($value, self::NORMALISE_CHARS);
+
+        // remove MBE suffix
+        $value = preg_replace('/ (mbe|m b e)$/i', '', $value);
+
+        $value = preg_replace('/([^a-z0-9-])/i', '', $value);
```
gregtyler: NB: this allows hyphens, whereas `CasRec::normaliseSurname` did not. Might be an issue when comparing with existing data.

src/AppBundle/v2/Registration/SelfRegistration/Factory/CasRecFactory.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @param LayDeputyshipDto $dto
+     * @return CasRec
+     */
+    public function createFromDto(LayDeputyshipDto $dto)
+    {
+        $entity = new CasRec($this->convertDtoToArray($dto));
+        $entity->setUpdatedAt($this->dateProvider->getDateTime());
```
gregtyler: What's the benefit of using a provider over `new \DateTime()`? Abstraction?

src/AppBundle/v2/Registration/Uploader/LayDeputyshipUploader.php
```diff
@@ -0,0 +1,190 @@
...
+            ->clientIsAttachedButNotToThisDeputy($layDeputyshipDto->getCaseNumber(), $layDeputyshipDto->getDeputyNumber());
+
+        if (false === $result) { return false; }
+
+        // Some deputies have a ',' separated list of deputy nums so if the above query tells us the client is registered to
+        // a different deputy, it may be a false result if the deputy num is concealed with a string list.
+        // Double check for the deputy num within a string list.
+        $deputyNumbers = explode(',', $result['deputy_no']);
+
+        return (in_array($layDeputyshipDto->getDeputyNumber(), $deputyNumbers)) ? false : true;
```
gregtyler: Would this be simpler as `!in_array(...)` rather than ternary?

src/AppBundle/Entity/Repository/ClientRepository.php
gregtyler: I've just seen how this is used in `clientBelongsToDifferentDeputy`, and that the array is needed. In that case, I'd suggest changing the name to something like `getOtherAttachedDeputies` to make it clear what should come out.

aiesoftware: Good suggestion, I'll take another look at the naming of the method.

src/AppBundle/Service/DataNormaliser.php
aiesoftware: Well spotted. I added the ability to accept hyphens in the surname as it felt like a bug to not accept it. I was in two minds about altering the existing functionality though. More than happy to keep it as it was.

src/AppBundle/v2/Registration/SelfRegistration/Factory/CasRecFactory.php
aiesoftware: The benefit comes with unit tests, being able to mock the response when making assertions on the value of a date. If source code uses `$v->setFoo(new \DateTime())` and a tests asserts `$this->assertEquals(new \DateTime(), $v->getFoo())`, it can cause sporadic test failures as the value of the `seconds` can differ between the two objects.

src/AppBundle/v2/Registration/Uploader/LayDeputyshipUploader.php
aiesoftware: Sure, I'll update.

src/AppBundle/v2/Registration/SelfRegistration/Factory/CasRecFactory.php
aiesoftware: See usage here: https://github.com/ministryofjustice/opg-digi-deps-api/pull/566/files#diff-eebdd052d15cc0753febeefb8efc2c4c

gregtyler: Oh that's really smart, nice.

src/AppBundle/v2/Registration/Assembler/CasRecToLayDeputyshipDtoAssembler.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->normaliser = $normaliser;
+    }
+
+    /**
+     * @param array $data
+     * @return LayDeputyshipDto
+     */
+    public function assembleFromArray(array $data)
+    {
+        return
```
slizzio: Should there be a call to canAssemble here?

src/AppBundle/v2/Registration/Assembler/LayDeputyshipDtoCollectionAssembler.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function assembleFromArray(array $data)
+    {
+        $collection = new LayDeputyshipDtoCollection();
+
+        foreach ($data as $uploadRow) {
+
+            if (!$this->layDeputyshipDtoAssembler->canAssemble($uploadRow)) {
+                throw new \InvalidArgumentException('Cannot assemble LayDeputyshipDto: Missing expected data');
+            }
```
slizzio: could move this to the assemble method

aiesoftware: I could but I felt that making `canAssemble` a part of the public interface would enforce any other assemblers (i.e a Sirius one) to implement the same check.

aiesoftware: On reflection, it does seem sensible to encapsulate it within the service, to prevent the call being bypassed. I'll update.

PR #568, opened 2019-07-17
DDPB-2751 Ensured email address is lowercase before requesting new token

slizzio:
## Purpose
Ensure password reset works with both upper and lower case input.

Fixes [DDPB-2751](https://opgtransform.atlassian.net/browse/DDPB-2751)

## Approach
strtolower email in User entity when requesting email, in case any database entries are in upper case, and changed api call to reset token with lower case email

## Checklist
- [ ] I have updated documentation (Confluence/GitHub wiki) where relevant
- [ ] I have added tests to prove my work, and they follow our [best practices](https://github.com/ministryofjustice/opg-digi-deps-client/wiki/Testing-best-practices)
- [ ] I have successfully built my branch to a feature environment
- [ ] There are no Composer security issues (`docker-compose run api php app/console security:check`)

### DIFF
--- a/src/AppBundle/Controller/UserController.php
+++ b/src/AppBundle/Controller/UserController.php
@@ -312,7 +312,7 @@ class UserController extends RestController
         }
 
         /** @var EntityDir\User $user */
-        $user = $this->findEntityBy(EntityDir\User::class, ['email' => $email]);
+        $user = $this->findEntityBy(EntityDir\User::class, ['email' => strtolower($email)]);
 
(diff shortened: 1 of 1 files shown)


aiesoftware:
As mentioned in the ticket, do we not want to lowercase all emails at the point that they are persisted?

slizzio:
Can confirm that all emails that use the User entity are in lowercase. the api User entity has         
$this->email = strtolower($email); checked and everything should and is going through the entity. There are no instances of SQL that add users without using the entity. I have however added it to a few other places (v2 DTOs mainly)

PR #570, opened 2019-07-23
DDPB-2786: Clone assets and bank accounts after resubmission

gregtyler:
## Purpose
When a report is submitted, the next year's report is automatically created. DigiDeps carries across the assets and bank accounts from the old report so they don't have to be re-added.

When the report is then unsubmitted and resubmitted, any changes to the assets and bank accounts are _not_ carried across. This means deputies have to add that information twice.

Fixes [DDPB-2786](https://opgtransform.atlassian.net/browse/DDPB-2786)

## Approach
I separated out the process which clones assets and bank accounts into a standalone function (`ReportService::clonePersistentResources`). I then updated `ReportService::submit` so that it calls the function when a report is resubmitted (as well as maintaining the functionality when it's initially submitted).

In order to stop duplication, I've added a check to identify if the bank account or asset already exists. Only strictly new resources will be copied across.

## Learning
This was a good exercise in writing a solid unit test. I tided up some existing tests too, so that the new function was tested in isolation and mocked out elsewhere. We should probably do the same for `createNextYearReport` in the future.

## Checklist
- [x] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant
- [ ] I have successfully built my branch to a feature environment


src/AppBundle/Service/ReportService.php
```diff
@@ -104,18 +104,67 @@ private function createNextYearReport(Report $oldReport)
...
-        foreach ($oldReport->getAssets() as $asset) {
-            $newAsset = clone $asset;
-            $newAsset->setReport($newReport);
-            $this->_em->detach($newAsset);
-            $this->_em->persist($newAsset);
+        $toReport->setNoAssetToAdd($fromReport->getNoAssetToAdd());
+        foreach ($fromReport->getAssets() as $asset) {
+            // Check that the target report doesn't already have a matching asset
+            $assetExists = false;
+            foreach ($toReport->getAssets() as $toAsset) {
```
aiesoftware: Out of curiosity - are these existence checks for the scenario of a resubmitted report? I'm assuming in the case of a first submit, the new report will have no assets and never meet this condition?

gregtyler: Yep. The genesis of this work is that we need to copy the assets/accounts across when resubmitting reports, but we don't want to duplicate things that were created on first submit or mess with data that's been changed since.

So it will only clone the records across if they're definitely new, and won't edit or delete anything.

aiesoftware: LGTM

PR #573, opened 2019-07-25
Ddpb 2724 bypass case used exception

slizzio:
## Purpose
To allow a lay deputy to register for a client when that client already exists under a professional deputyship.

Fixes [DDPB-2724](https://opgtransform.atlassian.net/browse/DDPB-2724)

## Approach
After much deliberation I have made the client soft deletable. This is to avoid removing the client completely and any reports that might be associated. So the action of discharging a client is to make them soft deleted. This prevents the client appearing anywhere in client. However admin area may sttill want to access the client (and reports) from the old professional deputyship so for admin pages I have disabled the soft delete filtter and re-enabled it once the query has been run. This allows the client to appear twice in the event that a lay deputy has re-registered against the case number. To clearly identify which is which I have applied a discharged date where required on the admin pages.

## Learning
_Any tips and tricks, blog posts or tools which helped you. Plus anything notable you've discovered about DigiDeps_

## Checklist
- [ ] I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant
- [ ] New and existing unit tests pass locally with my changes (`docker-compose run --rm api sh scripts/apiunittest.sh`)
- [ ] There are no Composer security issues (`docker-compose run api php app/console security:check`)


src/AppBundle/Service/UserRegistrationService.php
```diff
@@ -45,7 +45,7 @@ public function selfRegisterUser(SelfRegisterData $selfRegisterData)
         $existingClient = $this->em->getRepository('AppBundle\Entity\Client')->findOneByCaseNumber(CasRec::normaliseCaseNumber($selfRegisterData->getCaseNumber()));
 
         // ward off non-fee-paying codeps trying to self-register
-        if ($isMultiDeputyCase && $existingClient instanceof Client) {
+        if ($isMultiDeputyCase && ($existingClient instanceof Client) && $existingClient->hasDeputies()) {
```
gregtyler: The extra brackets here seem unnecessary?

gregtyler (dismissed):

src/AppBundle/Service/UserRegistrationService.php
slizzio: Force of habit, I always put each condition in their own set of brackets in case anyone adds an or condition and a bug is introduced. Not necessary with all &&'s

gregtyler:
[Transferred to opg-digideps](https://github.com/ministryofjustice/opg-digideps/pull/1)

